PR #3, opened 2023-10-29
Add support for Improv-based provisioning

thenewwazoo:
This change adds support for Improv-based wifi provisioning. Further configuration (e.g. for displaying a HomeKit QR code, etc) is still TODO.

NOTE that this code causes irregular watchdog timeouts on my test hardware, but I _strongly_ suspect the problem is in this hardware. I have successfully tested a full provisioning cycle with it, though, so I'm confident it's good enough for early development.


thenewwazoo:
@PaulWieland not sure how much insight/oversight/review you want, or if there's anyone else I/we should loop in for review.

PaulWieland:
@thenewwazoo I'm really limited on time at the moment, but I will test as soon as I can. In the mean time you can merge the PR if you like. Otherwise I'll can do it if you are ready.

thenewwazoo:
sgtm, I imagine you're probably slammed right now. :) I won't block on your input, but will continue to open PRs to give an opportunity for review.

PR #4, opened 2023-11-04
Initial support for garage door RX comms

thenewwazoo:
This commit adds initial support (basically, a port of the ratgdo
codebase) for Security+ 2.0 comms in RX mode. TX is TBD.

This code is still untested on real hardware, but there are unit tests for the logic.


.gitmodules
PaulWieland, line 1: Maybe best to reference our [forked version](https://github.com/ratgdo/secplus)

PR #29, opened 2023-12-13
Add a motion sensor service.

jgstroud:
When we get the motion detect message, notify HomeKit and start a pseudo timer.  When our timer reaches 0, send the motion clear

### DIFF
--- a/src/comms.cpp
+++ b/src/comms.cpp
@@ -121,6 +121,17 @@ void comms_loop() {
 
                         break;
                     }
+                case PacketCommand::Motion:
+                    {
+                        RINFO("Motion Detected");
+                        /* When we get the motion detect message, notify HomeKit. Motion sensor
+                           will continue to send motion messages every 5s until motion stops.
+                           set a timer for 5 seconds to disable motion after the last message */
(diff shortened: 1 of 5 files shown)


jgstroud:
Looks like ESP Home uses a timer callback to clear the motion after 3 seconds.  I wanted to keep the code as light as possible and I just have a counter that I decrement each time through the main loop.  When it gets to 0, we clear the motion event.

thenewwazoo:
If the preferred solution is indeed timer-based, let's make it a real timer, using `millis` and a pre-defined value to flip the bit. The amount of time between each invocation of the main loop is variable.

That being said, is a timeout-based solution best? I haven't played with the motion sensor. Is there a second message when the motion sensor resets, or the like?

jgstroud:
Yeah, I realize this implementation would would be quite variable.  I'll switch it to a timer.  There only seems to be a message indicating motion detected and not an additional message to indicate motion clear.

jgstroud:
Updated to use timer to clear.  motion shows detected in HomeKit until the sensor stops sending motion messages

thenewwazoo (approved):
Fantastic! Thank you so much for this PR

PR #31, opened 2023-12-13
Enable adding multiple homekit-ratgdo's to HomeKit

Avi0n:
It wasn't possible to add more than one homekit-ratgdo device to HomeKit because `HOMEKIT_CHARACTERISTIC(NAME)` was not unique. This change adds a unique name based on `ESP.getChipID()`

### DIFF
--- a/src/homekit.cpp
+++ b/src/homekit.cpp
@@ -5,6 +5,10 @@
 #include "ratgdo.h"
 #include "comms.h"
 #include "log.h"
+#include <ESP8266WiFi.h>
+
+#define DEVICE_NAME_SIZE 19
+#define SERIAL_NAME_SIZE 18
 
 // Bring in config and characteristics defined in homekit_decl.c
 extern "C" homekit_server_config_t config;
@@ -23,6 +27,12 @@ void target_door_state_set(const homekit_value_t new_value);
 homekit_value_t obstruction_detected_get();
 homekit_value_t active_state_get();
 
+// Make device_name available
+extern "C" char device_name[DEVICE_NAME_SIZE];
+
+// Make serial_number available
+extern "C" char serial_number[SERIAL_NAME_SIZE];
+
 /********************************** MAIN LOOP CODE *****************************************/
 
 void homekit_loop() {
@@ -30,6 +40,9 @@ void homekit_loop() {
 }
 
 void setup_homekit() {
+    snprintf(device_name, DEVICE_NAME_SIZE, "Garage Door %06X", ESP.getChipId());
+    String macAddress = WiFi.macAddress();
+    snprintf(serial_number, SERIAL_NAME_SIZE, "%s", macAddress.c_str());
 
     current_door_state.getter = current_door_state_get;
     target_door_state.getter = target_door_state_get;
--- a/src/homekit_decl.c
+++ b/src/homekit_decl.c
@@ -24,6 +24,9 @@ void identify(homekit_value_t _value) {
     printf("accessory identify\n");
 }
 
+char device_name[19];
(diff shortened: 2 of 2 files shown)


src/homekit.cpp
```diff
@@ -23,13 +23,17 @@ void target_door_state_set(const homekit_value_t new_value);
...
+extern "C" char device_name[16];
+
 /********************************** MAIN LOOP CODE *****************************************/
 
 void homekit_loop() {
     arduino_homekit_loop();
 }
 
 void setup_homekit() {
+    sprintf(device_name, "ratgdo %X\0", ESP.getChipId());
```
thenewwazoo, line 36: Since the physical thing being controlled is a garage door, let's keep the name `Garage Door XXXXXXXX`.

Also, consider fully-specifying the `%X` so leading zeroes don't get eaten and the size is bounded.

src/homekit_decl.c
```diff
@@ -51,7 +53,7 @@ homekit_characteristic_t obstruction_detected = HOMEKIT_CHARACTERISTIC_(
 homekit_accessory_t *accessories[] = {
     HOMEKIT_ACCESSORY(.id=1, .category=homekit_accessory_category_garage, .services=(homekit_service_t*[]){
             HOMEKIT_SERVICE(ACCESSORY_INFORMATION, .characteristics=(homekit_characteristic_t*[]){
-                    HOMEKIT_CHARACTERISTIC(NAME, "Garage Door"),
+                    HOMEKIT_CHARACTERISTIC(NAME, device_name),
                     HOMEKIT_CHARACTERISTIC(MANUFACTURER, "ratCloud llc"),
                     HOMEKIT_CHARACTERISTIC(SERIAL_NUMBER, "123456"),
```
thenewwazoo, line 58: Should this also be set to the chip ID?

Avi0n: It should already be, `device_name` is getting the chip ID from `homekit.cpp`

src/homekit.cpp
Avi0n: Changed to "Garage Door XXXXXX" and limited to 6 hex ints. Initially I thought `.getChipID()` would return the entire MAC address, but turns out [it only returns the last 3 bytes](https://github.com/esp8266/Arduino/issues/921).

src/homekit_decl.c
thenewwazoo: sorry, I was referring to the serial number

Avi0n: Ah gotcha. Added chipId to `SERIAL_NUMBER`.

src/homekit_decl.c
```diff
@@ -24,6 +24,9 @@ void identify(homekit_value_t _value) {
     printf("accessory identify\n");
 }
 
+char device_name[19] = "XXXXXX\0";
```
thenewwazoo, line 27: The rvalue (`XXX...`) isn't necessary here, as bss section variables are initialized to zero, and the values are immediately overwritten. You can just define them (`char device_name[19];`) uninitialized and it's fine.

src/homekit.cpp
```diff
@@ -23,13 +23,21 @@ void target_door_state_set(const homekit_value_t new_value);
 homekit_value_t obstruction_detected_get();
 homekit_value_t active_state_get();
 
+// Make device_name available
+extern "C" char device_name[19];
```
thenewwazoo, line 27: consider making the numbers 19 and 7 into `const usize` variables or `#define`s, since they appear in multiple places.

donavanbecker (approved):

PR #34, opened 2023-12-13
Use WiFi.macAddress

donavanbecker:
extension of #31


donavanbecker:
#31 takes care of all of this.

PR #40, opened 2023-12-14
Add pin based obstruction detection.

jgstroud:
Pulled almost verbatim from the mqtt-ratgdo firmware.  Updated to more closely match the style of this project.


donavanbecker (approved):
Just tested this and it is working awesome!

Issue #25, opened 2023-12-09
Set activity LED off unless there is activity


adrianradtke:
LED stays on with current version. Set LED to default off unless there is activity (ie command sent from HomeKit, door in operation)

PaulWieland:
fwiw, the mqtt firmware turns the led on when searching for wifi. Once connected it shuts off. It then flashes whenever it receives a command.

thenewwazoo:
Fixed in #43

PR #43, opened 2023-12-16
Add LED activity

Avi0n:
Turn off LED for 500ms upon receiving transmit.
Addresses #25

### DIFF
--- a/src/comms.cpp
+++ b/src/comms.cpp
@@ -30,6 +30,8 @@ SecPlus2Reader reader;
 uint32_t id_code = 0;
 uint32_t rolling_code = 0;
 
+extern long unsigned int led_on_time;
+
 /*************************** FORWARD DECLARATIONS ******************************/
 
 void sync();
@@ -235,6 +237,10 @@ void comms_loop() {
 /********************************** CONTROLLER CODE *****************************************/
 
 bool transmit(PacketAction& pkt_ac) {
+    // Turn off LED
+    digitalWrite(LED_BUILTIN, HIGH);
+    led_on_time = millis() + 500;
+
     // inverted logic, so this pulls the bus low to assert it
     digitalWrite(UART_TX_PIN, HIGH);
     delayMicroseconds(1300);
--- a/src/ratgdo.cpp
+++ b/src/ratgdo.cpp
@@ -13,15 +13,18 @@
 void setup_pins();
 void obstructionLoop();
 void IRAM_ATTR isr_obstruction();
+void led_loop();
 
 /********************************* RUNTIME STORAGE *****************************************/
 
 struct obstruction_sensor_t {
-    unsigned int low_count = 0;  // count obstruction low pulses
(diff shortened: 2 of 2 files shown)


src/comms.cpp
```diff
@@ -38,6 +38,7 @@ uint32_t read_counter_from_flash(const char* filename);
 bool transmit(PacketAction& pkt_ac);
 void door_command(DoorAction action);
 void send_get_status();
+void blink(bool trigger);
```
jgstroud, line 41: I think this is function no longer exists

src/comms.cpp
```diff
@@ -232,9 +233,20 @@ void comms_loop() {
     }
 }
 
+// Check if it's been long enough to turn on the LED
+void led_loop() {
+    if (digitalRead(LED_BUILTIN) && millis() > garage_door.led_on_time) {
+        digitalWrite(LED_BUILTIN, LOW);
+    }
+}
+
```
jgstroud, line 242: I added a late comment on your other commit.  I was thinking maybe we could change this to a serviceTimerLoop() and move the code from https://github.com/ratgdo/homekit-ratgdo/blob/v0.6.0/src/comms.cpp#L217 here

jgstroud:
This iteration looking a lot better.

src/comms.cpp
thenewwazoo: consider putting this in ratgdo.cpp since it's not related to transmission or receipt of data _per se_

src/ratgdo.h
```diff
@@ -64,6 +64,7 @@ struct GarageDoor {
     bool light;
     LockCurrentState current_lock;
     LockTargetState target_lock;
+    long unsigned int led_on_time;
```
thenewwazoo, line 67: since this isn't a property of the garage door, I'd prefer it live at the top level (i.e. as a free variable in ratgdo.cpp)

thenewwazoo (changes requested):

src/ratgdo.h
jgstroud: LoL.... he had it defined as an independent global with externs.  I suggested moving it here though I was wondering if that was wrong based on it not actually being part of the garage door.

src/comms.h
```diff
@@ -12,4 +12,6 @@ void close_door();
 
 void set_lock(uint8_t value);
 void set_light(bool value);
+
+void led_loop();
```
jgstroud, line 16: This shouldn't go in comms.h.  led_loop() is defined in ratgdo.cpp, and is only referenced in ratgdo.cpp, so just move the declaration to the top of ratgdo.cpp

If you needed to call it from another cpp file, then you would put this in the ratgdo.h file

jgstroud:
This iteration looks good to me.  I still think I would change the led_loop() to a service_timers_loop() and move all timer servicing here.  I can make that change though as I have another formatting correction that needs to go in

thenewwazoo (approved):

PR #44, opened 2023-12-16
Get version from manifest file and QR code from svg. fixes #38 and #45

### LINKED ISSUE #38: auto_firmware_version.py not safe/reliable
The` auto_firmware_version.py` script is not safe way to establish the version.  The script presumes that` git describe --tags` will work, but that requires that the git repository actually have tags.  Without it you get:
```
fatal: No names found, cannot describe anything.
```
And the firmware version embedded into the source is empty.  This matters because anyone who forks the repository will not carry over all the tags, each repository have their own tag history.

A better place to pull the version number from would be `manifest.json` file.  This contains the actual version number and is independent of any git tagging.

dkerr64:


donavanbecker:
With this change the Home App actually displays the FirmwareRevision now:
image

donavanbecker:
Also QR looks good also:
Screenshot 2023-12-17 at 9 48 58 AM

thenewwazoo:
I apologize for the brusqueness with which I closed the QR code issue before. I wasn't aware that no further comments would be permitted (I think that's a configurable that isn't available to me). I very much appreciate you doing the work to improve how versioning works for contributors working in their own repos, and for submitting a good solution to including the QR code in the repo. Now that I'm aware of the actual flash size (and have [corrected the record](https://github.com/ratgdo/homekit-ratgdo/commit/d76d7af52572ba02329337d9983196e62c1d7986)), I'm fine with including it.

My only concern is that one of the first things a new developer sees in the web.cpp file is a big block of raw SVG markup. Can you move that into a header file to hide it a bit?

thenewwazoo:
The build and release process is still in flux. The process encoded in [`x.sh`](https://github.com/ratgdo/homekit-ratgdo/blob/main/x.sh) was, more than anything, something that worked, was _slightly_ less manual, and that I didn't have to think too hard about. @donavanbecker has offered to look into github actions, which should be the ultimate solution.

Unfortunately, this PR as written breaks `x.sh release`. I think simply re-ordering the steps in `x.sh` will be the lowest-effort solution that maintains the status quo.

dkerr64:
> My only concern is that one of the first things a new developer sees in the web.cpp file is a big block of raw SVG markup. Can you move that into a header file to hide it a bit?

Yup, I can look into that.  Thanks.

dkerr64:
> > My only concern is that one of the first things a new developer sees in the web.cpp file is a big block of raw SVG markup. Can you move that into a header file to hide it a bit?
> 
> Yup, I can look into that. Thanks.

So this is another Wow! moment.  Thank you for prompting me to look into this. The result is way simpler inclusion of the QR code.  And I learned all about [raw string literals](https://www.stroustrup.com/C++11FAQ.html#raw-strings) thanks to [this](https://stackoverflow.com/questions/410980/include-a-text-file-in-a-c-program-as-a-char) discussion.

jgstroud:
It looks to me like you need to rebase your branch, and then I think you need to squash your QR code commits.  And unless I am missing something, it looks like qrcode.h didn't get committed.

dkerr64:
> The build and release process is still in flux. The process encoded in [`x.sh`](https://github.com/ratgdo/homekit-ratgdo/blob/main/x.sh) was, more than anything, something that worked, was _slightly_ less manual, and that I didn't have to think too hard about. @donavanbecker has offered to look into github actions, which should be the ultimate solution.
> 
> Unfortunately, this PR as written breaks `x.sh release`. I think simply re-ordering the steps in `x.sh` will be the lowest-effort solution that maintains the status quo but I obviously welcome any improvements you are willing to contribute.

In what way is `x.sh` broken?  I see that it is this script that moves the `.bin` file into place.  And I see that it is opening up the vim editor for `manifest.json`... but I don't see what it does to this file... are you saying that tat the manifest file is updated _after_ the build completes?

dkerr64:
> It looks to me like you need to rebase your branch, and then I think you need to squash your QR code commits. And unless I am missing something, it looks like qrcode.h didn't get committed.

oops, yes.  Just pushed it now.

jgstroud:
you merged main not your development branch.  do a `git rebase main`
I doubt @thenewwazoo is going to merge with broken commits.  You need to squash the 3 QR code commits.
Not sure about having the 2 unrelated commits in the same PR.

dkerr64:
> you merged main not your development branch. do a `git rebase main` I doubt @thenewwazoo is going to merge with broken commits. You need to squash the 3 QR code commits. Not sure about having the 2 unrelated commits in the same PR.

Okay, cleaned up now.  I went with brute force method, but net-net is the same, two clean commits.  Yes both in the same PR if that becomes an issue then I can split out, but more work for me... if both will be accepted then accept both at once.

jgstroud:
Much better.  Yeah, force push is the only way after you squash.  That's normal.
Also, very cool about the raw string literals.  I didn't know about this either.  But I only ever write C, so no surprise there.

PR #48, opened 2023-12-17
Remote syslog.- DO NOT MERGE

jgstroud:
Setup a syslog server that accepts UDP messages on port 514  and make sure you have a DNS record 'syslog' pointing to it.

Logging works great.  You don't see the messages from HomeKit library yet.  Unfortunately this caused my GDO to go unresponsive in HK although I could still control it.


jgstroud:
I have a version that logs all the HomeKit stuff too.  Will reopen after I clean it up.

PR #49, opened 2023-12-17
Send a get status packet after toggling the lights

DenisPalmans:
The local light state is not updated when turning on/off via HomeKit, this causes subsequent external light events (from keypad or motion) to be missed and not reported to Homekit.


thenewwazoo:
Nice catch, and thank you!

Issue #51, opened 2023-12-17
Refactor timers


thenewwazoo:
There are currently two timer-esque things (activity LED and motion sensor) that are handled alongside in the comms loop. While not representing a huge amount of computation time, it'd still be better if these were pulled into their own file or section of `ratgdo.cpp` (per [this comment](https://github.com/ratgdo/homekit-ratgdo/pull/43#issuecomment-1858913030))

thenewwazoo:
Fixed in #52

PR #52, opened 2023-12-18
Service all timers in one place

jgstroud:
Service all timers in one place.  The Obstruction loop was essentially just a timer service as well.
Fixes issue https://github.com/ratgdo/homekit-ratgdo/issues/51


thenewwazoo:
Another great contribution, thank you :)

Issue #36, opened 2023-12-14
Wifi times out


klipschthx:
I can't get the unit to connect to wifi. Ive tried it two different wireless networks, and it always times out. Any ideas on why that might be? Thx.

eric-ooi:
I'm seeing the same.  I can continuously ping the ESPHome firmware with low latency and no packet drops but from the same location I can't do the same with the HomeKit firmware.

eric-ooi:
I should mention that I get the same WiFi timeout message, but I can see it does connect to my network.  Once it gets an IP, doing the ping shows the latency and packet loss issues.

klipschthx:
Yeah I guess i should add that I know its not the hardware or the network because when i flash it with esphome it connects no problem.

jgstroud:
The issues with getting through the initial configuration may be fixed in 0.6.0 released today. I think the Wifi latency and packet loss is different than any issues related to getting connected for the first time.

klipschthx:
i installed 0.6.0 and same issue :(

klipschthx:
And I guess to be more specific, the issue is that the wifi times out upon connection. Another interesting/weird thing is that there are maybe 3-4 of the same SSIDs. Maybe they're different broadcast bands?

Avi0n:
> Another interesting/weird thing is that there are maybe 3-4 of the same SSIDs. Maybe they're different broadcast bands?

They're likely multiple APs broadcasting the same SSID (like mesh nodes). It seems other projects have sorted by signal strength, then removed all but the strongest from the list. I somewhat doubt this is related to wifi latency/packet loss though.

a1224:
FWIW, I am getting the timeout, but the device connects and I am able to continue to "Visit Device" by just pressing Skip.

eric-ooi:
Just flashed 0.6.0 and while I still get the initial timeout when trying to connect, it does in fact connect to my network (as before) but this time it appears that I can maintain a continuous ping without dropping packets and without high latency. 🤞

klipschthx:
you're totally right. it just times out on initial connect. good catch yall. hopefully it all works ok now.

eric-ooi:
Yep, I just wired mine up and all functions are quick to respond and work as described.  Awesome work, y'all!!

dmkw01:
I’d like to add my thanks as well. Thank you to Paul for the hardware and to all the others for contributing to the HomeKit coding.

eric-ooi:
[...]

PR #53, opened 2023-12-18
Remove duplicate SSIDs from setup web page

Avi0n:
If there are multiple APs broadcasting the same SSID (such as mesh setups), duplicate SSIDs show up in the WiFi setup dropdown. This PR fixes this by only displaying the network that has the strongest signal.
There's no directly associated issue, but [this comment ](https://github.com/ratgdo/homekit-ratgdo/issues/36#issuecomment-1858448789) mentions the problem.

### DIFF
--- a/src/wifi.cpp
+++ b/src/wifi.cpp
@@ -159,16 +159,46 @@ bool on_command_callback(improv::ImprovCommand cmd) {
 void get_available_wifi_networks() {
     int networkNum = WiFi.scanNetworks();
 
+    int sortedIndicies[networkNum];
+    for (int i = 0; i < networkNum; i++) {
+        sortedIndicies[i] = i;
+    }
+
+    // sort networks by RSSI, strongest to weakest
+    for (int i = 0; i < networkNum; i++) {
+        for (int j = i + 1; j < networkNum; j++) {
+            if (WiFi.RSSI(sortedIndicies[j]) > WiFi.RSSI(sortedIndicies[i])) {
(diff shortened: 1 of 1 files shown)


Avi0n:
Here's a slightly different version using `std::sort` instead of `swap` if preferred.
https://github.com/Avi0n/homekit-ratgdo/commit/ccf3b38b1b9352b3e0ff2ae2948122e8306794d2

thenewwazoo:
It looks like the esphome implementation does [something similar](https://github.com/esphome/esphome/blob/0a117eb562204a28d1955aa8eb6d2963c625d95a/esphome/components/improv_serial/improv_serial_component.cpp#L223) but without the RSSI preference. Do you see the same thing?

Avi0n:
I hadn't seen esphome's implementation, but yes that's what I'm seeing as well.

I'd like to keep the RSSI sorting for cases where the strong network gets cut, because the device could have a hard time connecting if the signal is weak. Most networks with multiple APs have Client Steering enabled, but if they don't, the device would continually try to connect to the weaker AP due to storing the BSSID of the weaker AP it initially connected to.

dkerr64:
Is it possible to allow user to manually enter SSID name?  If a SSID name is not broadcast then it will not show up.  Some network SSIDs are deliberately hidden from broadcasting but can still be connected to if you enter it manually.

jgstroud:
That's certainly a very useful feature, but I think it's unrelated to this PR. Should open a separate issue for that

PR #54, opened 2023-12-19
Build Using GitHub Actions

donavanbecker:
This is an example of the Build running and then attaching the firmware.bin file to the release: [donavanbecker/homekit-ratgdo-v0.7.0](https://github.com/donavanbecker/homekit-ratgdo/releases/tag/v0.7.0)

This is an example of the firmware.bin and firmware.elf being attached to the workflow run:
[Build Workflow](https://github.com/donavanbecker/homekit-ratgdo/actions/runs/7261739296)

A few approaches that we could use depending on how you want to handle this.

Currently I have it designed to only trigger on a create of a release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,80 @@
+name: Build Firmware - PlatformIO CI
+
+on:
+  repository_dispatch:
+    types: [build-firmware]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Tag
+        id: tag
+        uses: JinoArch/get-latest-tag@latest
+
+      - uses: actions/checkout@v3
+        with:
+            submodules: secplus
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/pip
+            ~/.platformio/.cache
+          key: ${{ runner.os }}-pio
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+      - name: Install PlatformIO Core
+        run: |
+            pip install --upgrade pip
+            pip install --upgrade platformio
+
+      - name: Install External Dependencies
+        run: |
+          cd lib
+          git clone https://github.com/argilo/secplus.git
+
+      - name: Build PlatformIO Project
+        run: pio run -e ratgdo_esp8266_hV25
+
+      - name: Rename Firmware
+        run: |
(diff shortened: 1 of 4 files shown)


BEFH:
Any reason why this uses MacOS instead of Linux?

thenewwazoo:
Can you add a `release.md` file in the `docs/` dir documenting how to perform a release? Also, what's the interaction with this and the changes in #44? I suspect that change could (or should) get glommed onto this one.

donavanbecker:
It should be Linux.

@thenewwazoo, I actually have more updates for this, where we actually update the manifest.json file also. Give me a little bit and I will get that added to this PR, still testing

donavanbecker:
> Can you add a `release.md` file in the `docs/` dir documenting how to perform a release? Also, what's the interaction with this and the changes in #44? I suspect that change could (or should) get glommed onto this one.

I think you can merge that one and then this one and should be good.

I have actually been testing this with that PR merged into my fork.

donavanbecker:
Okay @thenewwazoo, this PR is ready. I have create a `releasing.md` file in the docs folder. I have made this really easy. Literally all you have to do is create a Github Release/tag and it will do the rest. 

See commits that are created from the action when updating files: [homekit-ratgdo:commits](https://github.com/donavanbecker/homekit-ratgdo/commits/main/)
See Release assets: [v0.10.0](https://github.com/donavanbecker/homekit-ratgdo/releases/tag/v0.10.0)

donavanbecker:
I was just following what you had so far. We can set in the workflow that it must follow the v[0-9].[0-9].[0-9] patterned. Would you like me too?

.github/workflows/build.yml
```diff
@@ -0,0 +1,80 @@
...
+          python-version: '3.9'
+      - name: Install PlatformIO Core
+        run: |
+            pip install --upgrade pip
+            pip install --upgrade platformio
+
+      - name: Install External Dependencies
+        run: |
+          cd lib
+          git clone https://github.com/argilo/secplus.git
```
radford, line 36: We shouldn't need to do this if the `submodules: secplus` works above.

donavanbecker: When I tried it without this, it didn't work. So, I included this.

radford: I think you need `submodules: true` instead of `submodule: secplus`.

donavanbecker: nice! I correct that with #57.

PR #50, opened 2023-12-17
Add HTTPUpdateServer OTA support

sstoiana:
Apparently all of the ESP8266 modules on the ratgdo are 4m and not 1m flash. Should address #20.

Issue #45, opened 2023-12-16
Embed the QR code into the web page HTML


dkerr64:
Well, I don't appreciate closing issue #42 without any opportunity to discuss.  And as GitHub would not let me reopen it, I am opening a new issue.

I understand the need to keep the firmware image as small as possible, but offloading HTML (or in this case the QR code PNG) to an off-site server is not the answer.  The incremental size is small (the PNG is way larger than it need be) and the QR code can easily be embedded into the HTML as SVG... see [here](https://github.com/dkerr64/homekit-ratgdo/commit/1ee53f189a32a09b01b69d872036bd2823b6d05c).

Yes, I understand that internet connectivity is required to flash the device.  But that does not mean that it should be required for setup.

I have not opened a PR with the above commit only because if you shutdown #42 without discussion, why bother if you won't merge it.

dkerr64:
Humm, I just realized that by pushing to the same branch as PR #44 then merging that would pull in this too.  Sorry, that was not my intention. You can cherrypick the manifest.json version number change if you just want that and not this.

But... firmware image size is a bigger problem than the size of the HTML for the web page.

PR #55, opened 2023-12-19
Update Description for Each Action Button

### LINKED ISSUE #38: auto_firmware_version.py not safe/reliable

donavanbecker:
Wait till these are merged
[Remove duplicate SSIDs from setup web page #53](https://github.com/ratgdo/homekit-ratgdo/pull/53)
[Add HTTPUpdateServer OTA support #50](https://github.com/ratgdo/homekit-ratgdo/pull/50)
[Get version from manifest file and QR code from svg. fixes #38 and #45 #44](https://github.com/ratgdo/homekit-ratgdo/pull/44)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 .piolibdeps
 compile_commands.json
 __pycache__
+.vscode/*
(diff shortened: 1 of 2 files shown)


donavanbecker:
Merged everything into #57

Issue #27, opened 2023-12-09
Infinite (maybe) loop when not connected to GDO


dkerr64:
I'm testing with just the board itself, not yet connected to garage door opener.  Successfully paired to HomeKit but if I then try and open the door it goes into a loop...

```
!!! [ 227921] RATGDO: transmit failed, will retry
>>> [ 227927] RATGDO: Collision detected, waiting to send packet
!!! [ 227933] RATGDO: transmit failed, will retry
>>> [ 227939] RATGDO: Collision detected, waiting to send packet
!!! [ 227944] RATGDO: transmit failed, will retry
>>> [ 227950] RATGDO: Collision detected, waiting to send packet
!!! [ 227956] RATGDO: transmit failed, will retry
>>> [ 227963] RATGDO: Collision detected, waiting to send packet
!!! [ 227969] RATGDO: transmit failed, will retry
>>> [ 227975] RATGDO: Collision detected, waiting to send packet
!!! [ 227980] RATGDO: transmit failed, will retry
>>> [ 227986] RATGDO: Collision detected, waiting to send packet
!!! [ 227992] RATGDO: transmit failed, will retry
>>> [ 227998] RATGDO: Collision detected, waiting to send packet
!!! [ 228004] RATGDO: transmit failed, will retry
>>> [ 228010] RATGDO: Collision detected, waiting to send packet
!!! [ 228016] RATGDO: transmit failed, will retry
>>> [ 228022] RATGDO: Collision detected, waiting to send packet
!!! [ 228027] RATGDO: transmit failed, will retry
>>> [ 228033] RATGDO: Collision detected, waiting to send packet
!!! [ 228039] RATGDO: transmit failed, will retry
>>> [ 228045] RATGDO: Collision detected, waiting to send packet
!!! [ 228051] RATGDO: transmit failed, will retry
>>> [ 228057] RATGDO: Collision detected, waiting to send packet
!!! [ 228063] RATGDO: transmit failed, will retry
>>> [ 228069] RATGDO: Collision detected, waiting to send packet
!!! [ 228075] RATGDO: transmit failed, will retry
>>> [ 228081] RATGDO: Collision detected, waiting to send packet
!!! [ 228087] RATGDO: transmit failed, will retry
>>> [ 228093] RATGDO: Collision detected, waiting to send packet
```

Eventually I think it crashes and reboots.  Connecting this to my GDO is not practical for testing purposes.

thenewwazoo:
[...]

PR #58, opened 2023-12-21
prevent continuous logging when transmit fails

kharamel:
Resolves https://github.com/ratgdo/homekit-ratgdo/issues/27

I only tested this using Serial Monitor. Didn't seem to need implement 2nd option on https://github.com/ratgdo/homekit-ratgdo/issues/27#issuecomment-1857603126

### DIFF
--- a/src/comms.cpp
+++ b/src/comms.cpp
@@ -41,6 +41,7 @@ bool transmit(PacketAction& pkt_ac);
 void door_command(DoorAction action);
 void send_get_status();
 
+
 /********************************** MAIN LOOP CODE *****************************************/
 
 void setup_comms() {
@@ -220,12 +221,15 @@ void comms_loop() {
 
     } else {
         PacketAction pkt_ac;
+        static bool logged_transmit_failed_err = false;
 
         if (q_peek(&pkt_q, &pkt_ac)) {
             if (transmit(pkt_ac)) {
                 q_drop(&pkt_q);
-            } else {
+                logged_transmit_failed_err = false;
+            } else if (!logged_transmit_failed_err){ 
                 RERROR("transmit failed, will retry");
+                logged_transmit_failed_err = true;
             }
         }
     }
@@ -245,8 +249,12 @@ bool transmit(PacketAction& pkt_ac) {
     delayMicroseconds(130);
 
     // check to see if anyone else is continuing to assert the bus after we have released it
+    static bool logged_collision_err = false;
     if (digitalRead(UART_RX_PIN)) {
-        RINFO("Collision detected, waiting to send packet");
+        if (!logged_collision_err) {
+            RINFO("Collision detected, waiting to send packet");
+            logged_collision_err = true;
+        }
         return false;
     } else {
         uint8_t buf[SECPLUS2_CODE_LEN];
@@ -265,6 +273,7 @@ bool transmit(PacketAction& pkt_ac) {
         }
     }
 
+    logged_collision_err = false;
     return true;
 }
 


SausCode:
I believe at startup prev and current will likely be the same value, which means it will not update the status. I think on the first call to this function you will always want to force an update. Say for instance door is obstructed, esp loses power, obstruction cleared, power restored, status is not updated correctly.

jgstroud:
Why are we adding obstruction status in the comms module?

jgstroud:
I think you are confusing transmit collisions with obstructions.

jgstroud:
To resolve #27 you need to remove all the obstruction stuff you added and you just need a single static variable to track the previous collision state and output the log message on collision change.

kharamel:
Made modifications based on feedback above.

jgstroud:
Can you edit the name of this PR and squash your commits?

jgstroud:
The log_coll_error variable seems to use inverse logic.  It's true when there is no error. You should invert it

kharamel:
log_coll_error indicates we want to log error. We don't want to do that unless we have had successful transmission since we last logged the error. Based on that, don't we want what's implemented here?

edrikk:
The name of the log_coll_* variables are not really good in my opinion.
Once/if this PR is accepted the concept of "collision" will be lost and people will have to read the code to see what it means .. "log collect error" maybe?

I would suggest being a bit more verbose in this case with the name, and given that it's a true / false even add _ind to them.
e.g. log_collision_err_ind and log_collision_info_ind

jgstroud:
I definitely agree.  The variable names are odd and have nothing to do with the collision.  It's also confusing logic to set the error flag true after a successful transmit. 

I'm also not sure we really need the 2 different log messages and variables.

kharamel:
In my opinion given the context, coll in variable name was enough to indicate collision. However, I've update the variable name to be more verbose. Also, the variable was not an error flag but a flag that indicated if the error needed to be logged. I've made change for that to be more clear by changing it to indicate if the error was logged. From my understanding, we retry to transmit only from comms_loop. I think It makes sense to log `...will retry` there. Since comms_loop is not the only one calling transmit, having a single variable to handle logging in both places would cause interference.

jgstroud:
This reads so much better.  I think it's ok like this.  Also, you misunderstood my other comment, or I just wasn't clear.  I think it would just be more simple if we remove the RERROR("transmit failed" message all together along with the new variable and just leave the collision detected message in the transmit function.  Not sure what good it does to have the 2 messages, especially when we don't log an error for the other calls to the transmit function.  Also removes some extra unnecessary instructions.

kharamel:
That makes sense. I didn't want to discount any reasoning the author might have had when implementing it. @thenewwazoo any objections to getting rid of this [line](https://github.com/ratgdo/homekit-ratgdo/blob/main/src/comms.cpp#L228)?

thenewwazoo:
If I'm understanding correctly, the concern is that in a certain improper configuration (namely, a ratgdo that's unconnected to a garage door opener), the UX is bad. I'm receptive to that, but I think this is PR slightly misses (or doesn't clearly define) the desired outcome in that case, and perhaps conflates that with normal collisions.

Bus contention is normal and expected, and having its occurrence logged is (probably) a useful thing when diagnosing problems in a properly-installed ratgdo. I'd prefer to define how to detect error conditions that result in repeated bus collisions, and then act on that. Improving the UX in the case of improper configuration/installation would fall out naturally as a result.

I don't think exponential backoff makes sense because multiple subsequent collisions definitely indicates a problem we don't want to work around (or that we expect to resolve itself). An initial thought is setting an error-condition status bit and refusing to transmit again, but I'm open to suggestions.

jgstroud:
I just loaded up a new ratgdo and it's sitting on my bench not connected to anything and I do not see this logging loop.  Seems like we are solving sort of a Don't Care scenario?  Now I'm thinking more about this and as collisions will happen under normal operation, but they should not be continuously happening back to back, so having this log message printing every time we have a collision would indicate that something is clearly wrong.

I think maybe we just leave this PR alone and leave the issue as working as designed.   If we decide to log to flash, then we'll need to revisit this or rate limit it somehow.

What are your thoughts @thenewwazoo ?

jgstroud:
As I've been messing around with testing on a unit not connected, I have hit this issue.  But for that benchtop testing, I just remove the check and let it transmit the packet.  In normal operation this shouldn't be happening, and the continuous log messages are an indicator that something is broken in your setup.  I think we should just close this without merging and close the issue as working as designed.

PR #73, opened 2024-01-02
Fix a few duplicate log messages

### LINKED ISSUE #61: log message for door state and lock state are identical
Current lock state log message [here](https://github.com/ratgdo/homekit-ratgdo/blob/2b590057b7081276f0b00bce471a3109112a5aca/src/homekit.cpp#L112) is the same as current door message [here](https://github.com/ratgdo/homekit-ratgdo/blob/2b590057b7081276f0b00bce471a3109112a5aca/src/homekit.cpp#L73)

Same for target states [here](https://github.com/ratgdo/homekit-ratgdo/blob/2b590057b7081276f0b00bce471a3109112a5aca/src/homekit.cpp#L118) and [here](https://github.com/ratgdo/homekit-ratgdo/blob/2b590057b7081276f0b00bce471a3109112a5aca/src/homekit.cpp#L79) 

So you cannot tell in the log which state is being requested.

tabacco:
A couple of log messages relating to the lock referenced the door instead. This fixes those up.

Fixes #61

### DIFF
--- a/src/homekit.cpp
+++ b/src/homekit.cpp
@@ -109,7 +109,7 @@ homekit_value_t active_state_get() {
 }
 
 homekit_value_t current_lock_state_get() {
-    RINFO("get current door state: %d", garage_door.current_lock);
+    RINFO("get current lock state: %d", garage_door.current_lock);
 
     return HOMEKIT_UINT8_CPP(garage_door.current_lock);
 }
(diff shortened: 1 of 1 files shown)


jgstroud:
Would you mind squashing these 2 commits?

Issue #64, opened 2023-12-26
Offer a version of the firmware that has the motion detection compiled out


SausCode:
My gdo doesn’t have a motion sensor on it, so it is useless for me and just clutters up the home app. 

Would be great to have a way to opt out of this, either just ifdef’ing it out and offering two firmwares, or perhaps a config in the webpage that allows you to turn it off. 

There’s probably a similar use case to remove the light control, but I imagine most people’s openers have a light bulb built in.

SausCode:
Thinking about it more would likely be best done at the same time as WiFi config.

tabacco:
It seems reasonably easy to not register the motion detector accessory, and to expose a toggle in the web UI. The bit I don't know how to do is to store a preference flag somewhere, since that seems to be delegated to the homekit library right now (for pairing info), and there's no existing config support to piggyback off of.

PaulWieland:
This can't be added to the wifi config UI because ESP Web Tools specifically doesn't support any configuration options other than wifi credentials.

The Web UI can be configured to save preferences but requires the web service to store the configuration on the flash filesystem. This configuration then has to be parsed and read into memory at startup.

All of this adds a lot of complexity and overhead to the code base. It might be better to just offer two targets that are compiled using flags, one with and one without the motion sensor.

jgstroud:
I had considered just waiting until we see a motion message before advertising the motion service, but wasn't sure if HomeKit would properly handle the new service getting added after the fact.  It would be worth experimenting with.  It seemed pretty unobtrusive having the sensor available, so I didn't spend much time on that.

reh8388:
When connecting the device to HomeKit it gives you the option to select which attributes to add so you could deselect the motion sensor during setup. Doesn’t seem like something that needs to be tackled with the FW. 

![IMG_0619](https://github.com/ratgdo/homekit-ratgdo/assets/26860977/c48a2491-a4eb-4305-8951-61579a89438d)

dkerr64:
Once we pull in #63 we will have opportunity to add configurable settings through web page.  Right now nothing is implemented in the proposed web page design for this, but I think it could be added quite easily.

thenewwazoo:
HomeKit handles adding services on the fly badly; it doesn’t work, or didn’t the last time I tried it. 

[...]

PR #85, opened 2024-01-05
make the motion sensor dynamic.

jgstroud:
Only enable the service if a motion sensor is detected. By default the service will be disabled.  On the first motion detect message, set a flag to enable the service and reboot to trigger the addition of the new service in HK.  Will be permanently enabled after.

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -1,0 +1,41 @@
+// Copyright 2023 Brandon Matthews <thenewwazoo@optimaltour.us>
+// All rights reserved. GPLv3 License
+
+#include "utilities.h"
+#include "log.h"
+#include "LittleFS.h"
+
+
+void sync_and_restart() {
+    LittleFS.end();
+    delay(10);
+    ESP.restart();
+}
+
+uint32_t read_file_from_flash(const char* filename) {
+
+    File file = LittleFS.open(filename, "r");
+
+    if (!file) {
+        RINFO("%s doesn't exist. creating...", filename);
+
+        uint32_t count = 0;
+        write_file_to_flash(filename, &count);
+        return 0;
+    }
+
+    uint32_t counter = file.parseInt();
+
+    file.close();
+
+    return counter;
+}
+
+void write_file_to_flash(const char *filename, uint32_t* counter) {
+    File file = LittleFS.open(filename, "w");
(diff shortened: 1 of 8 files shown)


jgstroud:
Note, this change will almost certainly require you to repair after loading this for the first time.

jgstroud:
Fixes https://github.com/ratgdo/homekit-ratgdo/issues/64

jgstroud:
After you load this firmware, your device will show up with no motion sensor.  If you have a motion sensor, then after pairing your device with HK, trigger the motion sensor.  A short time later (it can take a minute), the motion device will show up in HK.

dkerr64:
Given that the Arduino HomeKit stack is buggy (It regularly and randomly un-pairs for me), I think it wise to hold off merging this until we have the Espressif HomeKit stack... which will hopefully be more reliable.

My concern is the need to go and activate the motion detector every re-pair (which for me is too often).

Also, Apple Home deletes any automations attached to an accessory if that accessory is removed.  So we want this to be a once-and-done thing.

Or maybe I'm missing something... once a motion detector is found, is that saved for good in EEPROM (like WiFi SSID/PW) or filesystem such that un-pair and re-pair will not require a new discovery of the motion detector?

donavanbecker:
@dkerr64 Why not merge and release updates until that point?

donavanbecker (approved):

dkerr64:
> @dkerr64 Why not merge and release updates until that point?

I'm not against merging updates.  I'm concerned that I will have to manually trigger motion to re-discover the motion detector every time I have to re-pair.  If it is a once-and-done then I'm happy.

jgstroud:
It's saved in flash.  So it's a one time activation.  Don't have to reenabled after repairing.

donavanbecker (approved):

src/utilities.cpp
```diff
@@ -0,0 +1,41 @@
+// Copyright 2023 Brandon Matthews <thenewwazoo@optimaltour.us>
+// All rights reserved. GPLv3 License
+
+#include "utilities.h"
+#include "log.h"
+#include "LittleFS.h"
+
+
+void sync_and_restart() {
```
thenewwazoo, line 9: This function calls the GDO Security+ 2.0 "sync" operation to mind. maybe "graceful_restart"?

PR #86, opened 2024-01-09
Update Release Process

donavanbecker:
Merged Manifest and Build Workflows into 1 Release Workflow

Here is a release example from my fork: [v1.0.0](https://github.com/donavanbecker/homekit-ratgdo-esp/releases/tag/v1.0.0)
Here is the run of the action: [v1.0.0-Workflow Run](https://github.com/donavanbecker/homekit-ratgdo-esp/actions/runs/7517282409/job/20463275705)


.github/workflows/manifest.yml
```diff
@@ -89,4 +89,5 @@ jobs:
       - name: Repository Dispatch
         uses: peter-evans/repository-dispatch@v2
         with:
-          event-type: build-firmware
\ No newline at end of file
+          event-type: build-firmware
+          client-payload: '{"tag": "${{ steps.tag.outputs.latestTag }}"}'
```
thenewwazoo, line 93: Is this still valid?

donavanbecker: Yes, that is the value that is being pulled by this file and pushed over to the build file. I guess I could just combine the two files if you want?

donavanbecker: I have merged the files into one release workflow action. So no need for this anymore.

PR #99, opened 2024-01-29
Discord Webhook after Release

donavanbecker:
This will run the discord webhook after the release happens. (Merges the two files together)

@thenewwazoo 
@jgstroud


jgstroud:
Think you might have pushed this commit to the wrong branch.  There are 4 commits, most of which seem to be related to the issue templates.

donavanbecker:
see #101

PR #113, opened 2024-02-12
Complete browser cache support, improve cosmetics of unpair

dkerr64:
Completes support for browser-side caching of CSS/JS/Images.
Updates the cosmetics of un-pair/reset HomeKit and reboots the device.


jgstroud:
@dkerr64 let me know if or when you are ready to merge and I'll do a quick test build and merge.

dkerr64:
Hi @jgstroud the only thing left (that I can think of) is to investigate if I can have the server notify the browser of status changes.  Right now the browser polls the server every 10 seconds asking for updates to status (door open, paired, etc).  This is hardly best practice.

However I will have to investigate if this can be done and I haven't started that.

In the meantime, this PR provides a couple of needed updates... better unpair cosmetics and reboot as part of that, and adding paired status to the 10-second-polling.  So I do think worthwhile merging.

Thanks

rayr007:
To send updates to the browser from the server, you could use WebSockets and stream messages back. Or possibly use this mechanism: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

dkerr64:
> To send updates to the browser from the server, you could use WebSockets and stream messages back. Or possibly use this mechanism: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

yes, I found example of server sent events here..
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WebServer/examples/ServerSentEvents/ServerSentEvents.ino and this looks to be fairly simple.

There are also examples for WebSockets but I suspect it might be too heavyweight and use more memory than necessary.

jgstroud:
> > To send updates to the browser from the server, you could use WebSockets and stream messages back. Or possibly use this mechanism: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events
> 
> yes, I found example of server sent events here.. https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WebServer/examples/ServerSentEvents/ServerSentEvents.ino and this looks to be fairly simple.
> 
> There are also examples for WebSockets but I suspect it might be too heavyweight and use more memory than necessary.

This seems like overkill.  The web status is nice, but it's more diagnostic than anything and not having it realtime isn't a big deal IMHO.

dkerr64:
I am closing this in favor of PR #117

PR #136, opened 2024-03-27
Add wifi RSSI to web page

jgstroud:


jgstroud:
@dkerr64 added wifi RSSI to webpage

src/www/index.html
```diff
@@ -72,9 +72,13 @@
               <td id="netmask"></td>
             </tr>
             <tr>
-              <td style="text-align: right;">Gatweway IP:</td>
+              <td style="text-align: right;">Gateweway IP:</td>
```
dkerr64, line 75: typo here in word Gateway.

jgstroud: lol.  I made the original typo worse

jgstroud:
I'm not sure I trust this value.  Both of my devices are reporting -52 and one is right next to the AP and other is much further and has a wall in the way.

dkerr64:
I can check what mine report, but want to wait to see if I can capture log of a crash/reboot before changing the firmware on the one connected to serial monitor.  My Unifi controller reports signal strength so I can compare two ratgdo's.  Right now one is reporting -63dBm and the other -46dBm.  But that is the strength the Unifi access point is seeing (curiously, each is connected to a different access point).

jgstroud:
On the AP side, unifi is reporting -53 and-57 for mine.  So, the one with the wall in the way is a bit lower, but not significant.  So the device RSSI values may be right.  I just dont know.  I do want your opinion on 2 things.  I just put it at the bottom of the list on the web page, but it probably makes more logical sense to move it right below the WiFi SSID line.  The other thing was its calling the function to get the RSSI value in the heartbeat.  I dont know if that makes sense or not.  Assuming you have one client open, then it's reading that value once a second, which is probably fine.  What do you think?

jgstroud:
It's also possible that my one device may be too close to the AP and saturating the receiver.  It might improve by moving it further away.

dkerr64:
I think it is better to have the RSSI right below the SSID.  I had a look at WiFi.RSSI() and it calls a function wifi_station_get_rssi() which I cannot find the source for.  I wanted to check if it was lightweight, in which case doing it on heartbeat is not a problem.  Googling turns up some discussion... looks like it returns average signal strength from last few packets received.  Sounds benign enough.

jgstroud:
I think that wifi_station_get_rssi() is hidden in the proprietary binary source driver for the wifi controller.

PR #139, opened 2024-03-29
Improve web page stability

dkerr64:
Use a UUID instead of client IP address to identify each unique browser.  And improve the timeout logic on client side so that we don't have a continuously running interval timer, which can cause multiple checkStatus() calls to get queued up.


src/web.cpp
```diff
@@ -18,13 +18,15 @@
...
 
+#include "EspSaveCrash.h"
 #include <arduino_homekit_server.h>
 #include <ESP8266WebServer.h>
 #include <ESP8266HTTPUpdateServer.h>
 #include <Ticker.h>
 #include "log.h"
 #include "utilities.h"
 
+EspSaveCrash saveCrash(1408, 0x400);
```
jgstroud, line 29: I think you need to add the 3rd parameter "true".  I missed this on my first revision.

PR #142, opened 2024-04-03
Include Elf file

donavanbecker:
These changes will include the elf file in committing to the repo and to the release.

If the repo part isn't needed just let me know.


dkerr64:
Thank you.  @jgstroud was working on something similar so I'll defer to him to review this PR... but as we now save the crash dump (once PR #139 is merged) we do need to save the elf file as well as bin to allow us to decode any reported crashes.

PR #153, opened 2024-04-04
Suspend certain activity when update underway, including comms and ho…

dkerr64:
…mekit loops


jgstroud:
This look ok to me.  I did an initial test of it and it worked, but I haven't been able to break the OTA, so I'm not sure how well it addresses the issue, but it seems like it could only make things better.  If you feel like this has been adequately tested, we can merge it.

dkerr64:
I agree it can only make things better.  Feels like a sensible thing to do, so yes I support merging.

PR #166, opened 2024-05-01
Add check for pre-releases in firmware update dialog

dkerr64:
The check for a new release on the main home page will NOT include check for pre-releases.

To check for pre-release you must open the firmware update dialog, select the checkbox, then click on the check for update button.

This way users are not notified of pre-releases automatically, but if you are in-the-know, you can check for and install them.

### DIFF
--- a/src/comms.cpp
+++ b/src/comms.cpp
@@ -43,7 +43,7 @@ SecPlus2Reader reader;
 uint32_t id_code = 0;
 uint32_t rolling_code = 0;
 uint32_t last_saved_code = 0;
-#define MAX_CODES_WITHOUT_FLASH_WRITE 100
+#define MAX_CODES_WITHOUT_FLASH_WRITE 10
 
 /******************************* SECURITY 1.0 *********************************/
 
--- a/src/web.cpp
+++ b/src/web.cpp
@@ -986,11 +986,13 @@ void handle_update()
         // Error logged in _setUpdaterError
         server.send(400, F("text/plain"), _updaterError);
     }
+#ifdef CRC_CHECK
     else if (!ESP.checkFlashCRC())
     {
         // CRC error
         server.send(400, F("text/plain"), "Flash CRC check failed. Update aborted");
     }
+#endif
     else
     {
         RINFO("Upload complete, received firmware MD5: %s", Update.md5String().c_str());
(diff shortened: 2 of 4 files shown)


donavanbecker:
Could this be added to the setting page? And then it is just on everywhere when it is checked?

donavanbecker:
Also are we needing a new workflow so we can trigger pre-releases?

dkerr64:
> Could this be added to the setting page? And then it is just on everywhere when it is checked?

Yes it could. But do we want the main page to alert users to pre-releases?

Also, if I put it on the settings page then I need to have that setting saved at and retrieved from the server.  Not a big deal.  But we should decide on what the requirement is... I thought it was for testing purposes, rather than to announce and send out pre-releases/betas to all.

dkerr64:
> Also are we needing a new workflow so we can trigger pre-releases?

Outside my area of expertise.  But GitHub clearly supports it, and @thenewwazoo used the prerelease flag for early releases as I see it in the releases JSON returned.

donavanbecker:
It could just be turned off by default, and then if they want prereleases they can.

donavanbecker:
> > Also are we needing a new workflow so we can trigger pre-releases?
> 
> 
> 
> Outside my area of expertise.  But GitHub clearly supports it, and @thenewwazoo used the prerelease flag for early releases as I see it in the releases JSON returned. 

Yeah I can create another workflow for this.

jgstroud:
I'm not sure we are all on the same page here.  I don't think we need a new workflow.  All we need is for the ratgdo webUI to only look for builds tagged as releases.  We can have a checkbox in the settings tab or in the update UI that says to also allow checking for new pre-release builds which is disabled by default.  The whole idea here is so that I can do new builds and test them before everyone starts downloading and installing potentially broken builds like happened yesterday.

the one change I wanted was for the workflow to skip the discord hook on a build tagged a pre-release.

donavanbecker:
Okay @jgstroud, this is already in place so if we don't want it to publish for pre-release we can just remove that here:

https://github.com/ratgdo/homekit-ratgdo/blob/ccf6691a180b1aa544876223959132cd101c7036/.github/workflows/release.yml#L192-L214

PR #167, opened 2024-05-02
Remove the visibility check as it was causing issues.

jgstroud:
@dkerr64 This patch seem to be working

### DIFF
--- a/src/www/functions.js
+++ b/src/www/functions.js
@@ -218,10 +218,9 @@ async function checkVersion(progress) {
     }
     // make sure we have newest release first
     let prerelease = false;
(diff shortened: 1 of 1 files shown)


dkerr64 (commented):
I think it okay to remove that.  I checked the docs and it does warn that this function has limited availability, so may not work in all browsers.

https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility

can remove the setting to false and just have...

let prerelease =  document.getElementById("prerelease").checked;

jgstroud:
Ugh... I need to quit rushing.  I made the change you requested and didnt push it.

PR #175, opened 2024-05-06
Fixes and features for next release

dkerr64:


viewlog.sh
```diff
@@ -0,0 +1,3 @@
+UUID=$(uuidgen)
+URL=$(curl -s "http://${1}/rest/events/subscribe?id=${UUID}&log")
+curl -s -N "http://${1}${URL}?id=${UUID}" | sed -u -n '/event: logger/{n;p;}' | cut -c 7-
```
jgstroud, line 3: This is fantastic!

dkerr64: > This is fantastic!

Thought you might like this.   I tested on Ubuntu ant MacOS.

PR #180, opened 2024-05-21
Continue to send SSEheartbeats during firmware update

dkerr64:
During firmware uploads we had suspended Server Sent Events (SSE) heartbeat messages.  If uploading on a slow connection, the browser client will timeout, shutting it down. I had thought the timeout was 5 seconds, turns out to be 30 seconds.

Solution was not as simple as just removing the test for `updateUnderway` in `SSEheartbeat()` function.  Turns out that during file upload nothing else runs... The main loop is not called, `Ticker` functions are not run, etc.  This explains why we had problems in the past with firmware uploads failing... as the HomeKit server was not getting called to process any HomeKit messages coming in.  We solved that by shutting down the HomeKit server before starting the upload.

So... I had to explicitly call the `SSEheartbeat()` function periodically during firmware uploads.  I do it on 10% increments, we should comfortably trigger this in under 30 seconds.

### DIFF
--- a/src/web.cpp
+++ b/src/web.cpp
@@ -135,7 +135,6 @@ bool _authenticatedUpdate;
 bool updateUnderway = false;
 char firmwareMD5[36] = "";
 size_t firmwareSize = 0;
-WiFiClient *firmwareClient = NULL;
 bool flashCRC = true;
 
 // Common HTTP responses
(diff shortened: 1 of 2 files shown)


jgstroud:
worked for me.

PR #191, opened 2024-05-27
Fixes / features for next release.

dkerr64:
Currently included:

- Set flash chip frequency to 26Mhz in hope that improves OTA firmware update reliability.
- Handle HTTP_HEAD method and improve web server log messages
- Save message log on clean reboot for later viewing
- Save result of `checkFlashCRC()` on reboot / crash.
- Initialize door lock state to unknown, and door state shows as unknown on web page until door is properly activated (but unknown door state cannot be returned to HomeKit as not a supported state)


src/ratgdo.cpp
```diff
@@ -53,6 +55,9 @@ void setup() {
 
     setup_comms();
 
+    // We can set current lock state to unknown as HomeKit has value for that.
+    // But we can't do the same for door state as HomeKit has no value for that.
+    garage_door.current_lock = CURR_UNKNOWN;
```
jgstroud: I think we just need to send a status command on boot to get the proper states

platformio.ini
```diff
@@ -11,7 +11,9 @@
 [env]
 upload_speed = 921600
 monitor_speed = 115200 ; must remain at 115200 for improv
+;monitor_speed = 74880  ; hardware default
 board_build.filesystem = littlefs
+board_build.f_flash = 26000000L
```
jgstroud, line 16: I don't want to merge this unless we find it absolutely necessary.

dkerr64: I actually think that this is the most important change to get wide testing.  Reducing from 40Mhz to 26Mhz could be the key to avoiding flash CRC errors.  Both my systems crashed this week unrelated to firmware upload.  Just maybe slowing down the flash chip frequency could make things more reliable.

src/web.cpp
```diff
@@ -1077,7 +1077,8 @@ void handle_firmware_upload()
         {
             uint32_t maxSketchSpace = (ESP.getFreeSketchSpace() - 0x1000) & 0xFFFFF000;
```
jgstroud, line 1218: getFreeSketchSpace already starts at the first free erase sector.  Why was this math added?

src/web.cpp
```diff
@@ -1077,7 +1077,8 @@ void handle_firmware_upload()
...
             RINFO("Available space for upload: %lu", maxSketchSpace);
-            if (!Update.begin(maxSketchSpace, U_FLASH))
+            RINFO("Flash chip speed %d MHz", ESP.getFlashChipSpeed() / 1000000);
+            if (!Update.begin((firmwareSize > 0) ? firmwareSize : maxSketchSpace, U_FLASH))
```
jgstroud, line 1220: What made you decide to make this change to the if statement?

src/ratgdo.cpp
```diff
@@ -39,7 +39,11 @@ void setup() {
     LittleFS.begin();
 
     Serial.printf("\n"); // newline before we start
-    RINFO("RATGDO setup starting");
+    RINFO("Starting RATGDO Homekit version %s", AUTO_VERSION);
+    RINFO("%s", ESP.getFullVersion().c_str());
+    RINFO("Flash chip size 0x%X", ESP.getFlashChipSize());
+    RINFO("Flash chip mode 0x%X", ESP.getFlashChipMode());
+    RINFO("Flash chip speed 0x%X (%d MHz)", ESP.getFlashChipSpeed(), ESP.getFlashChipSpeed()/1000000);
```
jgstroud, line 46: Hah.... I aded all the same code to my working branch

src/web.cpp
```diff
@@ -320,6 +322,7 @@ void setup_web()
         HeapSelectIram ephemeral;
 #endif
         server.on("/update", HTTP_POST, handle_update, handle_firmware_upload);
+        server.on("/verify", HTTP_POST, handle_verify, handle_firmware_verify);
```
jgstroud, line 325: @dkerr64 is there a way to combine these into 1?

src/web.cpp
```diff
@@ -1057,7 +1060,7 @@ void handle_update()
         server.send_P(200, type_txt, PSTR("Update Success! Rebooting...\n"));
         delay(100);
         server.stop();
-        sync_and_restart();
+        //sync_and_restart();
```
jgstroud, line 1063: for now we are not rebooting after an upload.  this allows a second call to verify and then a manual call to the reboot api.

src/ratgdo.cpp
dkerr64: We can back this out if you figure out how to get the status during initialization (fyi, I moved it from ratgdo.cpp into homekit.cpp).

src/web.cpp
dkerr64: I think I copied from somewhere else.  But now that I have looked at how the Updater.begin() code is written it is unnecessary.

src/web.cpp
dkerr64: When originally implemented I was not sending firmware size or MD5 up to the server, so the `Update.begin()` was being passed the maxSketchSpace.  The Updater code then comes to an "unexpected end" when the data runs out before it has reached the "size" expected.  It works fine, but as we now know the firmware size, the correct thing to do is pass that size into `Update.begin()`.  I have the conditional just in case there is a scenario where we don't know it (there is no firm requirement that the MD5 / firmware size is send before the file upload).

src/web.cpp
dkerr64: btw... we can probably also pass in the firmware size and MD5 in the same query string... replacing the need to first call `setgdo()` with those values.

src/web.cpp
jgstroud: Item 1 is wrong for a lot of reasons.  2 copies to flash instead of 1, you still have to verify the first copy. The problem with 2, is if a write somehow corrupts a previous block that has already been verified.

Once we root cause the OTA failures (and other flash corruptions), we can change the verify to just read the contents of the flash and recheck the MD5.  There should be no reason to do this byte by byte comparison.  However, until we root cause, I want to keep it like this.  We can revert this before we merge.

dkerr64: okay, then we can clean things up... let me work on the javascript side of things so it will report progress to user for each step.  Would you like an automatic reboot after the verification, or positive action from the user (press a reboot-now button).

PR #193, opened 2024-05-31
Firmware verification working branch

dkerr64:


dkerr64:
I merged in the master branch from https://github.com/cfurter/Arduino-HomeKit-ESP8266

The changes to ardunio_homekit_server.cpp are largely cosmetic... fixes to spelling errors, but there is one likely significant... a change from boolean && to a bitwise & operator in an if statement... if (!(ch->permissions & homekit_permissions_notify))

The changes to storage.c are considerable.  You can see the commit history here... https://github.com/cfurter/Arduino-HomeKit-ESP8266/commits/master/ only the last 6 are by courter, he forked from another fork which I had earlier merged in.  I think I made a mistake forking from the original https://github.com/Mixiaoxiao/Arduino-HomeKit-ESP8266... I should probably have started out by forking from one of the other forks as I now have a rather messy commit history that essentially tried to get my branch in sync with changes others had made.

My other "non-development" ratgdo crashed with the same error, and I used the esptool verify_flash command to confirm the same overwriting.  I have now installed the version that is on github onto both my devices.  I did have to reset/re-pair both to HomeKit as the pairing was lost.  That is unfortunate but unless you can decipher all cfurter's changes to storage.c and figure out why, I think we may have to live with this.

This version also has the checkFlashCRC() call on every request for status.  So you will notice a slowdown in page refresh.  Hopefully we can remove this once we have confidence that we have really found the bug.

dkerr64:
**Documenting history....**

I added a checkFlashCRC call every time the browser checks for status.  This morning I opened up the browser and saw the message that checkFlashCRC failed, after only 13 hours.  So...

```
$ python3 -m esptool -b 921600 -p /dev/ttyUSB1 verify_flash --diff yes 0x0000 ~/firmware-0531-10pm.bin
esptool.py v4.7.0
Serial port /dev/ttyUSB1
Connecting....
Detecting chip type... Unsupported detection protocol, switching and trying again...
Connecting....
Detecting chip type... ESP8266
Chip is ESP8266EX
Features: WiFi
Crystal is 26MHz
MAC: 08:3a:8d:f9:ef:36
Uploading stub...
Running stub...
Stub running...
Changing baud rate to 921600
Changed.
Configuring flash size...
Verifying 0x91e40 (597568) bytes @ 0x00000000 in flash against /home/david/firmware-0531-10pm.bin...
-- verify FAILED: 1164 differences, first @ 0x00000080
   00000080 00 7c
   00000081 00 12
   00000083 00 60
   00000084 00 88
   00000085 00 dc
   00000087 00 40
   00000088 00 cd
   00000089 00 20
```

And then we found in homekit storage.c I find this code...

```
int homekit_storage_reset_pairing_data() {

    byte blank[sizeof(pairing_data_t) * MAX_PAIRINGS];
    bzero(blank,sizeof(blank));

    INFO("Formatting HomeKit storage at 0x%x", PAIRINGS_OFFSET);
    if (!spiflash_write(PAIRINGS_OFFSET, blank, sizeof(blank))) {
        ERROR("Failed to erase HomeKit pairing storage");
        return -1; // Fail case
    }
    return 0;
}
```

The value of PAIRINGS_OFFSET is 128 (0x80)
The sizeof(blank) is 1280 (0x500)

Thus... we discovered that our HomeKit branch should probably sync up with that found [here](https://github.com/cfurter/Arduino-HomeKit-ESP8266).

PR #208, opened 2024-06-27
Fixes to SEC1.0 when no DIGITAL wall panel connected

mitchjs:
- changed transmit of sec1 byte (affected digital wall panel emulation)
- changed rules for packet (command) transmit when in wall panel emulation
- tweaked log to console for wall plate emulation

### DIFF
--- a/src/comms.cpp
+++ b/src/comms.cpp
@@ -155,50 +155,62 @@ void save_rolling_code() {
 
 void wallPlate_Emulation() {
     
-	if (wallPanelDetected) return;
+    if (wallPanelDetected) return;
 
-	unsigned long currentMillis = millis();
-	static unsigned long lastRequestMillis = 0;
-	static bool emulateWallPanel = false;
-	static unsigned long serialDetected = 0;
-	static uint8_t stateIndex = 0;
+    unsigned long currentMillis = millis();
+    static unsigned long lastRequestMillis = 0;
+    static bool emulateWallPanel = false;
+    static unsigned long serialDetected = 0;
+    static uint8_t stateIndex = 0;
 
-	if (!serialDetected) {
-		if (sw_serial.available()) {
-			serialDetected = currentMillis;
-		}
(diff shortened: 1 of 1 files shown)


dkerr64:
Could you squash the two commits into one, then we can merge into main.  Thanks.

src/comms.cpp
```diff
@@ -174,29 +174,31 @@ void wallPlate_Emulation() {
     // wait up to 15 seconds to look for an existing wallplate or it could be booting, so need to wait
 	if (currentMillis - serialDetected < 15000 || wallplateBooting == true) {
 		if (currentMillis - lastRequestMillis > 1000) {
-			Serial.println("Looking for security+ 1.0 wall panel...");
+			Serial.println("Looking for security+ 1.0 DIGITAL wall panel...");
```
jgstroud: These log messages won't log over the network.  Should consider using the macros instead.

mitchjs: yea i know... could change, but it was that way before i touched the code for the sec 1 fix
let me change it, then @dkerr64  figure out how to squash commits :)

jgstroud:
Is this ready for merge?

mitchjs:
@dkerr64, @jgstroud 
should be good to go.... i was hoping more would test it, but it looks like just me and 1 other... oh well

jgstroud:
Most of the users seem to have Sec+2 and most won't test until there is a new official release published.  I'll merge it and do a 1.5.1 build.

mitchjs:
ok, i assume you mean 1.6.1
im in debate to replace my old ass GDO, with a new Jackshaft one.... nothing really wrong with what i got
its just old :)

jgstroud:
Yes, 1.6.1. The lift master jackshaft opener appears to use the same wall panel that I have, so this firmware should work with those as well.

jgstroud:
I have a small PR I was working on.  I'll do a build when I'm done with that.

PR #216, opened 2024-07-14
For 1.6.1 release

dkerr64:
PR to collect changes for next release.  Currently includes two updates...

- Change log message timestamps to unsigned long type
- Include Flash CRC and length in viewlog/crashlog data.


jgstroud:
since this is related to logging, lets merge the change I made to the characteristic logging in HomeKit as well.

jgstroud:
Cool.  I didn't even think to try monitoring fragmentation

dkerr64:
> Cool. I didn't even think to try monitoring fragmentation

I noticed that maximum block size could reduce and then jump back up again.  So I have also added keeping track of the smallest max block size, which is done in the web loop.

dkerr64:
The `ESP.getMaxFreeBlockSize()` function appears to be destructive (maybe?).  Calling this function causes communication with the GDO to breakdown.  Trying to turn on/off light, etc. fails.  Door state is unknown.  So I have commented out this code for now pending further investigation.

Two theories... 1) the GDO comms is saving data somewhere that is getting overwritten by the max free block calculation code, or 2) calculating the max free block is too time consuming and interfering with timing of comms between ratgdo and the GDO.

If it is (2) then taking the calculation outsize of the main loop might help... instead of doing it every iteration, run it on a timer every e.g. 10 seconds?

dkerr64:
I'm not sure what the problem is when I request max free block size inside the loop() but it was definitely interfering with GDO comms.  I moved memory monitoring out into its own "ticker" that runs once a second (I started with it in the existing heartbeat function, but that only runs if a browser client is open).  This also keeps track of the worst case fragmentation.

During one of my reboots I encountered the MDNS responder crash during boot... it happened before the memory monitor ticker had started, so crash log reports 65535 free heap (the default I set).

Anyway, this gives us a tool to monitor memory heap fragmentation.

Issue #217, opened 2024-07-15
Not connecting to Wifi


seanpopke:
I ordered two ratgdos last fall. I installed both and both were working correctly. I recently noticed that neither were showing up in Apple HomeKit (I had installed the native HomeKit firmware.) When trying to reboot both, the blue light flashed once, which seems to indicate that they had lost their firmware? 

I’ve taken one into my office and reflashed it with the latest HomeKit firmware (1.6). The flash seems to work, but when I try to connect to Wifi, it isn’t successful. The logs show:

[58915119] RATGDO: WiFi disconnected SSID: iot24, BSSID: 6a:d7:9a:24:fa:6a, Reason: 203

I’ve tried erasing before re-flashing, but no change. Nothing has changed on my network or wifi access points. I’m using a Unifi system with a dedicated 2.4Ghz IOT SSID.

Any thoughts?

dkerr64:
The reason code 203 maps to WIFI_DISCONNECT_REASON_ASSOC_FAIL.  The ratgdo should continue to try and reconnect.

 I suggest looking at your Unifi logs to see if it gives any clues.

jgstroud:
Do you have them locked to an AP?  A lot of users did this early on, but it seems to be causing issues lately.  I recommend not locking the devices.

jgstroud:
Also, did you enable PMF on your wifi network?

seanpopke:
Thanks for the responses so far. They are not locked to an AP and PMF is disabled. I've made no changes to the wireless network between when I first set them up and now.

marylegz90:
I, too, am having a similar issue. I've updated my 2.5 to 1.6.0 after finding it "No Response" in HomeKit. Took a bit and a few reboots, but finally got the firmware web installer to get it "connected" to wifi. I found that it still is unresponsive in HomeKit but does show to be up in my Unifi console (Dream Machine Pro SE running 4.0.6). It doesn't stay connected hardly at all though, very randomly will it even ping.

PMF and AP locking is turned off on my network. No multicast or IGMP snooping is enabled either - as I had seen someone else say that worked for them.
![image](https://github.com/user-attachments/assets/8624ef14-8393-402e-8326-936eee9ab573)

Logs every once and a while throws back "Reason: 4" followed by multiple "Reason:203s" before cycling back through.
![image](https://github.com/user-attachments/assets/cf5d8705-38f5-4bcf-bdb4-497bb2282365)

[...]

PR #244, opened 2024-10-27
In preparation for 1.8.1 release...

dkerr64:
### What's Changed

* New feature... Allow selection of time zone when NTP server enabled.
* Change... We use built in Arduino core NTP client for time instead of separate module
* Change... Replace DNS lookup with a ping to gateway to test for network connectivity
* Bugfix... Don't display that an update is available if running newer pre-release
* Bugfix... When changing SSID in soft access point, make sure the WiFi settings are reset to DHCP
* Bugfix... wifiSettingsChanged setting had been removed by mistake.

### Known Issues

* Same as v1.8.0...
* Occasional failure to connect to WiFi. Tracked in <https://github.com/ratgdo/homekit-ratgdo/issues/217>


jgstroud:
@dkerr64 Ok, so I changed the DNS lookup to a gateway ping check.  The old way was fine, but I wanted to eliminate the need to do an internet lookup.  With the DNS method, mine would lookup google.com because my DNS server is on a different subnet.  Also fixed a bug that was introduced when you refactored the settings.

dkerr64:
Do you think we should log every minute that we pinged the gateway ok?  Feels excessive.  Maybe only log if it fails and/or latency is excessive?

jgstroud:
yeah, I was wondering if that was going to be too noisy.  ill rework it

dkerr64:
Now that I have timezone implemented, I'm thinking maybe should use ratgdo time (if available) for syslog.

Also... for timezones, I have provided selection for usual USA timezones.  But I could load the full list from the CSV file here... https://github.com/nayarsystems/posix_tz_db

jgstroud:
could we autodetect the timezone on the client side?

jgstroud:
I think you can auto discover you timezone based on your public IP address using https://ip-api.com

jgstroud:
Maybe its too much to implement on the esp8266, but this will return your timezone
curl http://ip-api.com/

jgstroud:
@dkerr64 I pushed a function that will lookup the timezone offset from that url I sent earlier.  You can use it if you want.  If you dont want to use it, we can drop the commit

dkerr64:
@jgstroud Thanks for the timezone lookup.  We also need to convert from e.g. Americas/New_York into the posix encoding, in this example "EST5EDT,M3.2.0,M11.1.0".  I'm doing this now on the browser side with a hardcoded list, but the full list is available online (link above).

jgstroud:
Yeah I saw that.  I just used UTC<offset>.  That sets the time correctly, but doesn't display a friendly timezone when you print the time.  I think this works if you just remove the time zone when you print the time. 

Otherwise we can just use your browser selector.

jgstroud:
Maybe give my commit a try for auto setting the timezone.  If you don't like it we can remove the commit. My testing, it worked fine.  But, to handle daylight savings time, we would probably need to check the offset daily just after midnight.

dkerr64:
> Maybe give my commit a try for auto setting the timezone. If you don't like it we can remove the commit. My testing, it worked fine. But, to handle daylight savings time, we would probably need to check the offset daily just after midnight.

Your lookup is useful.  I think we can solve for the 2nd part... worst case in in the browser javascript side if not on the ratgdo itself.  I'll look into it this evening.

dkerr64:
Also, on the gateway pings, I see you do a `millis() % 60000 == 0` test.  I don't think this is reliable, it requires that the loop is executed exactly on a multiple of 60 seconds.  If it executes just one millisecond later then the test fails.  I think we need to set a static `timeout = mills() + 60000` and then test for `millis() > timeout`

Thanks.

jgstroud:
Yeah, you are right.  There is no way that would be reliable.  I've been writing Verilog recently where this wouldn't be an issue.  My context switching in my head just isn't what it used to be.  I'll make the change.  Thanks for catching that.

jgstroud:
@dkerr64 I fixed the mistake with the timers.  also, for the Timezone, I was just setting it to UTC<offset>, so I removed the timezone code from the timeString.  But I don't want to overstep.  If you want to change it, go right ahead.  I just got the itch to make the auto-timezone thing work.  My feelings won't be hurt if you completely redo that.

dkerr64:
Here is what I am working on for time zone...

1. Default value for timezone is blank.
2. If timezone blank, ratgdo requests timezone from external IP address, as you have coded.  If not blank then it is assumed that user has already set desired timezone so no need to request.
3. Maybe... if SSID is changed, reset timezone to blank?
4. If the browser receives a timezone that is region/city only and no POSIX (simple test for a separating semicolon) then the browser will lookup the posix code and send it to the server.
5. If cannot fine a POSIX encoding, then sent UTC to the server.
6. From that point on we're all good.

On the browser side... we will asynchronously load the timezone CSV from https://github.com/nayarsystems/posix_tz_db and use that to do the lookup.  We will also use it to populate a selection field in settings page.  If loading from that URL fails, then use built-in list of half-dozen USA timezones.

jgstroud:
That all sounds good.  The only suggestion might be in step 1, instead of having a blank selection, have a default selection of "Auto".

dkerr64:
Have implemented time zone.  If you switch SSID then timezone is reset and discovered from external IP address. I'm not sure if this is good idea or not, but it provides us a way to test it.

dkerr64:
I'm inclined to finish this up with update to readme for timezone, and then publish this.

Then maybe we look harder at dry contact support as several people are requesting it... before we get tied up with ESP32.

dkerr64:
> I agree. Let's not add any more to this merge. Yeah, there have been quite a few dry contact requests recently. My plan was to just lift the dry contact code verbatim from https://github.com/ratgdo/mqtt-ratgdo/blob/2.5/src/ratgdo.cpp
> 
> The biggest reason I haven't touched it yet is because I have no way to test it.

One of my doors has a 20-year old opener.  I have not connected a ratgdo to it, but it may well require dry contact.

PR #249, opened 2024-11-16
First attempt at adding dry contact control

tlhagan:
It didn't seem like there was much movement towards adding dry contact control so I thought I would give it a shot.
I took the sections from https://github.com/ratgdo/mqtt-ratgdo/blob/2.5/src/ratgdo.cpp that were relevant from what I could tell.

I'm sure I've missed things, I have experience in other languages but very little in C++ so I hope you'll be patient with me.

I have a dry contact GDO and have figured out how to build/flash the code locally so I should be able to test things.


dkerr64:
@tlhagan Thank you for stepping up to help.  A couple of suggestions...

This would be best added as its own separate source/header file rather than adding to _ratgdo.cpp_... but that can wait until you are comfortable that things are working well.

And I see use of IRAM_ATTR... I am not sure what the effect of that is on our use of IRAM for allocating some blocks of memory during startup, but I suggest you go into the _log.h_ file and change the line `#define LOG_BUFFER_SIZE 8192` such that the size is much smaller... say 1024.  We can put it back up once we know if IRAM_ATTR uses the same block of memory.

Thanks.

PaulWieland (commented):
Anything related to the light controls can be discarded as dry contact controlled doors generally use an analog signal to control the light and so it's not supported.

tlhagan:
Thank you for the guidance. I removed the light related code and lowered the buffer size. I also added the dryContactLoop function to the main loop, if that is not correct you can let me know.

Another question I had was about the controlProtocol variable. In the mqtt code I see it defaults to secplus2 and can be secplus1 or drycontact, but I couldn't find where it was first declared and how it it is changed from the default. Could you explain how it is working, and if we need it in the homekit code?

jgstroud:
The IRAM_ATTR's are telling it to put the Interrupt service routines in IRAM so it's not fetching them from flash so they can be serviced quickly and deterministically.  We would need to verify, but I don't think this will get put in the location we are using for the IRAM heap allocations.

src/ratgdo.cpp
```diff
@@ -174,6 +175,138 @@ void setup_pins()
 
 /*********************************** MODEL **************************************/
 
+/*************************** DRY CONTACT CONTROL OF LIGHT & DOOR ***************************/
+void IRAM_ATTR isrDebounce(const char *type){
```
jgstroud, line 178: This looks like. lot of code for an ISR.    It's too complex and even has string functions in it.  ISRs should be kept as simple as possible.  Consider just capturing the pin that caused the interrupt, the time and the state of the pin and do the rest outside of the ISR.


Alternatively, I've used the OneButton library on ESP devices and in my experience, it does a pretty good job at debouncing switches.  In my limited use of it, it works quite well.  It might be a bit easier and cleaner.
https://github.com/mathertel/OneButton

maybe worth a proof of concept.

tlhagan: Thanks for the suggestion. I spent some time learning about the OneButton library and took a stab at replacing the ISR/debounce code with OneButton functions. You can let me know if you think I'm doing anything wrong.

dkerr64:
> My next step will be to try and add the communication with HomeKit.

You should look at the file `comms.cpp` and search on _"notify_homekit"_ the available functions are in `homekit.h`.  The HomeKit part calls functions defined in `comms.h` to trigger door open/close.

In fact it is probably best to integrate your code into `comms.cpp`, look for where we test for `gdoSecurityType`, and add a new possible value to that and then add the dry contact code when this is set to whatever value you select (we currently use values 1 and 2.  You can take 0 or 3 or 99 or whatever you want, 0 or 3 would be best as easier to integrate with the user interface code).

src/ratgdo.cpp
tlhagan: I was confused, the description of the dry contact pins in ratgdo.h had me thinking they were for open/close wall buttons. Now I understand they are for the open/close limit switches and I fixed the OneButton functions accordingly. What I'm still unsure of is whether I need to add additional functions to update dryContactDoorOpen/Close to false when the switches are released.

dkerr64:
@tlhagan I just merged the tweaks I was working on into main, you will need to rebase your branch onto the new release.

Thanks.

tlhagan:
Sorry I'm also not very experienced with rebasing. I gave it a shot but I thought it was done incorrectly, so I force pushed back to the commit before the rebase (perhaps not the most proper solution). Now I tried again but the rebase results seem the same. The code seems ok, but is the commit history correct? From what I expected the latest commits from you/the bot should be before my commits on this branch, but they are after.

dkerr64:
@tlhagan okay, I looked at your fork and it looks like the rebase was done backwards... an attempt to apply my changes on top of your changes, when we want it the other way around.

I recommend that you reset the HEAD of your branch to the last commit on Nov 22... this will get you back to where you were (all commits on the 23rd were the changes I merged).  Then just continue working on it and when you feel it is ready to merge I will assist with rebasing.

It will help if you add me as a collaborator to your fork.

Thanks.

dkerr64:
@tlhagan I have cleaned up your repo...

I created a new branch "dry-contact" and moved all your commits onto that, then rebased it with the ratgdo main.
I reset your "main" branch to be exactly the same as ratgdo/homekit-ratgdo main branch.

This way you can keep your copy of "main" in sync with ratgdo's and do all your work on its own branch, dry-contact.  By having both the original main and your changes in two branches on the same repo it makes it easier to use GUI tools to visualize the changes and manage rebasing.  Personally I use Sourcetree from Atlassian.

That caused this PR to be closed.  So please open a new PR from the dry-contact branch.

Thanks.

PR #253, opened 2024-11-24
Adding dry contact control

tlhagan:


tlhagan:
@dkerr64 Thanks for the help with rebasing.

As an update, the current version compiles and flashes properly and allows for switching to dry contact protocol via the settings page (using a value of 3). The next test I performed was to wire up an open/closed limit switch. I am using just one switch with 3 terminals (ground, NC, and NO) to act as both switches (hopefully this shouldn't affect things). When I 'close the door' (bring close limit switch closed, open limit switch open) for the first time, HomeKit and the logs show a proper update to closed. But after that opening or closing the switches does not result in any more updates. This probably has to do with the OneButton functions I implemented, so I'll take a look at those next.

PaulWieland:
I haven’t reviewed the code yet, but just fyi - There needs to be logic to update HomeKit to opening and closing as well as open and closed. 

door opening and door closing is whenever both limit switches are open. 

To know which of the two you have to keep track of the last state of both switches. If open limit was closed and is now open, the door is closing, and vice versa.

This will be impossible to mimic with an no/nc switch unless you can somehow hold the switch between states so that both no and nc are open.

tlhagan:
That's a good point. The logic should be there in the dryContactLoop function to capture the opening/closing states, but I should test it. I do have two limit switches, I was just planning to wire one for convenience in my actual garage setup. I will go ahead and connect both of them for testing to see if opening/closing are properly captured.

@PaulWieland On a related note I did have a question for you. I took the dryContactLoop function from the mqtt code and it is currently basically unchanged in ratgdo.cpp here. I was confused about the logic where it is taking actions if the control protocol is not dry contact. Could you help me understand what is the purpose of those actions (print message, call open/close door function, set dryContactDoorOpen/Close to false)? It seems like I might be able to remove them here.

PaulWieland:
The dry contact loop handles the inputs on the control board for dry contact control over the door opener. There are inputs that let you trigger the door to open or close, or toggle the light on/off.

PaulWieland:
I guess to add further clarification… those inputs are re-purposed when you have a dry contact controlled door to read the limit switches. If you have a dry contact control door, you would never  need dry contact controls.

dkerr64:
FYI that #254 wants to use dry contact support to control a gate rather than garage door... and requests that we have option to remove the light service from the accessory.

dkerr64:
@tlhagan what is the status of this?  Is it working, ready to merge?  If so I'd like to squash all the commits into one and then do a code review.

Thanks.

tlhagan:
Sorry, I got busy with other things recently and wasn't able to work on this. The limit switch reading and door state detection is now working properly. I just need to wire it up to my GDO and test the commands. Give me a day or two and I will provide an update.

tlhagan:
Ok, it is working!!

One thing I wanted to note, I got a bit tripped up by the 3 terminal NO/NC switches I am using. When I was using the NO and COM terminals, I was expecting to detect a 'press' when the magnets were brought together but instead I was getting a 'release'. I solved this by switching to the NC and COM terminals, but if NO switches are supposed to be used (as I understood from the wiring diagram) and it's not me misunderstanding the wiring we would just need to invert the press/release OneButton functions.

Other than that, it is ready to merge from my end.

dkerr64:
@tlhagan Thanks.  I have just squashed all the commits into one to make it easier for me to review the changes.  You will need to use the `--rebase` flag next time you fetch and pull.

lib/ratgdo/log.h
```diff
@@ -21,7 +21,7 @@ void print_packet(uint8_t pkt[SECPLUS2_CODE_LEN]);
 // This can be large, but not too large.  IRAM heap is approx 18KB, we also need
 // space for other data in here, so during development monitor logs and adjust
 // this smaller if necessary.  IRAM malloc's are all done during startup.
-#define LOG_BUFFER_SIZE 8192
+#define LOG_BUFFER_SIZE 1024
```
dkerr64, line 24: Let's put this back to 8192 and test.  If it fails, lets try 4096, etc til we find one that works.

src/ratgdo.cpp
dkerr64, line 1: I'd like to move the dry contact code out of `ratgdo.cpp` / `.h` and into its own file.... like `drycontact.cpp` / `.h` I am considering doing the same for secplus1/2 so this would be a good start with dry contact.

src/ratgdo.cpp
```diff
@@ -155,17 +174,19 @@ void setup_pins()
...
+   
     pinMode(STATUS_DOOR_PIN, OUTPUT);
-    */
+    
     pinMode(STATUS_OBST_PIN, OUTPUT);
-    /*
+    
     pinMode(DRY_CONTACT_OPEN_PIN, INPUT_PULLUP);
     pinMode(DRY_CONTACT_CLOSE_PIN, INPUT_PULLUP);
-    pinMode(DRY_CONTACT_LIGHT_PIN, INPUT_PULLUP);
```
dkerr64, line 166: Are we not able to control the light with dry contact mode?

PaulWieland: no, light light controls are analog on most dry contact openers.

PaulWieland:
> NC and COM terminals
You are right that you are supposed to be using the NO Terminals.

Do you have  a multimeter that you can test your switches with? Maybe they are simply mislabeled? You could also just test them with the ESP Home firmware to ensue they are behaving consistently.

With no magnet near, the input pin is left floating. Bringing the magnet near will pull the input pin low.

src/ratgdo.cpp
dkerr64: so, @PaulWieland , it would make sense to remove the light service from HomeKit so that a switch that can't do anything is not shown in Apple Home.  Correct?  Only for dry-contact mode of course.

PaulWieland: Yes agreed. With dry contact control the only thing you can do is open and close the door (and get door status feedback)

tlhagan:
I tested and they are indeed mislabeled. I guess that's what happens when you order cheap switches from amazon. So the press/release functions are working properly.

lib/ratgdo/log.h
tlhagan: I tested with 8192, it seems to be working fine. When would you expect a fail if it were to happen?

src/ratgdo.cpp
tlhagan: That's fine with me, let me know if that's something you would like me to attempt or you will take care of.

100ideas:
@tlhagan @PaulWieland I also received mislabeled magnetic reed switches that came from amazon - NC and NO labels were swapped. Had to figure it out with a multimeter continuity mode.

Somehere someone (reddit? here?) had recommended this particular product for adding dry contact support to ratgdo, so that what I got. But they are very trashy. Just posting this for posterity and in case anyone else ends up troubleshooting these switches.

**NC and NO terminals are labeled incorrectly** 2X Magnetic Reed Switch Normally Open Closed NC NO Door Alarm Window Security/Magnetic Door Switch/Magnetic Contact Switch/Reed Switch, DC 5Vto12V Light (Black)
https://www.amazon.com/gp/product/B086GYJLML (vendor is TAKHRWOD)

lib/ratgdo/log.h
dkerr64: It would fail immediately during startup. So I think safe to put back.

src/ratgdo.cpp
dkerr64: Let's leave it as it is and I will separate it out in the future when I look at separating out the others.

Olivierbkk:
Paul deleted issue 259 although we were close to getting it working. May I know what a “dry contact” version means for you? Is it only for sensors or also for actuation?

dkerr64:
> Paul deleted issue 259 although we were close to getting it working. May I know what a “dry contact” version means for you? Is it only for sensors or also for actuation?

Yes, well it is his product, so if he wants to delete a discussion that had diverged from the main purpose of the product then he can.

Keep working on your use case.  You might consider forking the project.  If you find you need to make changes you can open a PR.

Olivierbkk:
Thanks for the encouragement, unfortunately I do not have the programming skills.

dkerr64:
> Thanks for the encouragement, unfortunately I do not have the programming skills.

Keep me informed of what you find.  Before 259 was deleted I think you said you were seeing multiple toggles.  I found something while doing the ESP32 version which may explain that.  So I need to make an update for that.

Olivierbkk:
yes it did a few open/close short motions then went forward or backward until reaching a sensor. I saw the relay clicking a few times. I only saw that behavior when connected to the door controller, when I was testing on my table I saw the relay clicking once only, it was apparently working correctly. That's why I think it could be a sensor signal debounce issue, or perhaps there could be a timer to delay any additional trigger commands after a trigger. Let me know if you need more details.

antongitt:
I installed version 1.8.4, but since I only have the Open limit switch wired, the current logic cannot reliably detect when the door is fully closed (or even opened).
Would it be possible to add functionality to support configurations where only a single limit switch is available?

For example, in the case where only the Open limit switch is wired:
1. If the Open limit switch is pressed, it triggers state = open, assuming the door is now open.
2. If the Open limit switch is released, the system assumes the door is closing.
3. Wait a Fixed Time (door closing duration) before assuming the door is fully closed.

This would help users who do not have both limit switches but still need accurate open/close state tracking.
Thanks for your great work!

PaulWieland:
Two limit switches are required in order to detect the four door states.

tlhagan:
> I installed version 1.8.4, but since I only have the Open limit switch wired, the current logic cannot reliably detect when the door is fully closed (or even opened). Would it be possible to add functionality to support configurations where only a single limit switch is available?
> 
> For example, in the case where only the Open limit switch is wired:
> 
>     1. If the Open limit switch is pressed, it triggers state = open, assuming the door is now open.
> 
>     2. If the Open limit switch is released, the system assumes the door is closing.
> 
>     3. Wait a Fixed Time (door closing duration) before assuming the door is fully closed.
> 
> 
> This would help users who do not have both limit switches but still need accurate open/close state tracking. Thanks for your great work!

If you only want to use one limit switch, I would suggest using a 3 terminal switch (COM/GND, NO, NC). If you have it located as an open switch (e.g. magnets are together when door is open), then you would wire to the NC terminal in place of the open switch and NO in place of the closed switch in the dry contact wiring diagram [here](https://ratcloud.llc/pages/wiring). If it is located as a closed switch, you would reverse that. The opening/closing states will essentially be skipped, but open and closed should be read properly.

Issue #261, opened 2024-12-22
Not handling millis() rollovers


jgstroud:
The code is so stable now that my millis() counter actually rolled rover.  
https://docs.arduino.cc/language-reference/en/functions/time/millis/

My uptime had mysteriously gone back to 0, which originally I thought there must have been a power flicker.   But there was nothing in the logs.  Then today I clicked the wall button and the device went into recovery mode after a single press.  I went back and reviewed my syslog logs and see that the counter reset timeout value was set before the rollover.

jameyers4:
I, for one, would appreciate this change incorporated into 1.8.3 at earliest reasonable convenience. I was closing in on 40 days when I had to power down for non RATDGO reasons - I was close (-ish) to the 32 bit counter limit…

jgstroud:
PR is available.  There is no way to artificially test this as far as I can tell.  I'm debating whether I should just merge it now or wait 50 days to verify it properly handles rollovers.

dkerr64:
Please test with pre-release v2.0.0 (select pre-release box on update dialog).  Closing this issue, please open a new issue if problems found.

PR #262, opened 2024-12-23
Add a 64bit version of the millis() function

jgstroud:
millis() will rollover after 50 days leading to some interesting bugs. Create a 64bit version.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -23,7 +23,7 @@ build_flags =
     -Llib/lwip2
     -llwip2-536
     ${env.build_flags}
-    '-Wno-unused-variable'
+    '-Wall'
     -D PIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY_LOW_FLASH
     -D PIO_FRAMEWORK_ARDUINO_MMU_CACHE16_IRAM48_SECHEAP_SHARED
     -D LOG_MSG_BUFFER
--- a/src/comms.cpp
+++ b/src/comms.cpp
@@ -52,8 +52,8 @@ uint32_t last_saved_code = 0;
 
 static const uint8_t RX_LENGTH = 2;
 typedef uint8_t RxPacket[RX_LENGTH * 4];
-unsigned long last_rx;
-unsigned long last_tx;
+uint64_t last_rx;
+uint64_t last_tx;
 
 bool wallplateBooting = false;
 bool wallPanelDetected = false;
@@ -190,10 +190,10 @@ void wallPlate_Emulation()
     if (wallPanelDetected)
         return;
 
-    unsigned long currentMillis = millis();
-    static unsigned long lastRequestMillis = 0;
+    uint64_t currentMillis = millis64();
+    static uint64_t lastRequestMillis = 0;
     static bool emulateWallPanel = false;
-    static unsigned long serialDetected = 0;
+    static uint64_t serialDetected = 0;
     static uint8_t stateIndex = 0;
 
     if (!serialDetected)
@@ -268,7 +268,7 @@ void comms_loop_sec1()
     if (sw_serial.available())
     {
         uint8_t ser_byte = sw_serial.read();
-        last_rx = millis();
+        last_rx = millis64();
 
         if (!reading_msg)
         {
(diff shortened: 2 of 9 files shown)


jgstroud:
Fixes #261

dkerr64:
Looks like the commit removed code related to syslog port.  Was that intentional?

jgstroud:
Nope.  Something must not have merged properly.  I'll rebase and update.

jgstroud:
@dkerr64 good catch.  fixed now.  that was strange.  its like part of a single commit was missing from my branch.  rebased again and now I think it looks right.
I'm going to go ahead and keep running with this until I verify the rollover works before I merge it.

mitchjs:
out of curiosity, what "interesting bugs" did a 59day rollover create?
as rollover of millis() is not usually problem (as long as everything is unsigned longs and subtraction used)

thanks
Mitch

jgstroud:
In particular I saw my device go into recovery mode when I clicked the wall button.  We have a timeout value that should have reset but the millis rolled and the timeout ended up being way off in the future

dkerr64:
> @dkerr64 good catch. fixed now. that was strange. its like part of a single commit was missing from my branch. rebased again and now I think it looks right. I'm going to go ahead and keep running with this until I verify the rollover works before I merge it.

Thanks, looks good now.

dkerr64:
@jgstroud I added this for the ratgdo32 version, you can see the commit [here](https://github.com/ratgdo/homekit-ratgdo32/pull/26/commits/9b65bf2c05470c622132aa7f5683c8483b340ec1).

There were quite a few `unsigned long` variables that I found needed to be change to long long  (or `uint64_t`) which are missed in your changes here.  I suggest a grep for `unsigned long` and/or a look at the changes I made for ratgdo32.

jgstroud:
I'll take a look but I made sure it compiled without warning.  I would have expected type mismatch warnings.

dkerr64:
> I'll take a look but I made sure it compiled without warning. I would have expected type mismatch warnings.

I googled and apparently it is a _feature_ of the C compilers to accept assignment to different types, and it will truncate if necessary.  You can have it warn with a `-Wconversion` flag.  So I tried that... bad idea, thousands of warnings from just about every source file (not just ours).

jgstroud:
Yeah, it's definitely legal in C.  I really thought it would throw a warning if you didnt explicitly cast the data type, but I think I'm wrong there.  I also forgot that a long is 32 bits on this device.  Pushed new code that hopefully catches the rest of the missed variables

dkerr64:
> Yeah, it's definitely legal in C. I really thought it would throw a warning if you didnt explicitly cast the data type, but I think I'm wrong there. I also forgot that a long is 32 bits on this device. Pushed new code that hopefully catches the rest of the missed variables

Yes, I think you got them all now.  Thanks.

dkerr64:
> > Yeah, it's definitely legal in C. I really thought it would throw a warning if you didnt explicitly cast the data type, but I think I'm wrong there. I also forgot that a long is 32 bits on this device. Pushed new code that hopefully catches the rest of the missed variables
> 
> Yes, I think you got them all now. Thanks.

Spoke too soon. I found another one... `lastDoorUpdateAt` in web.cpp needs to be declared 64-bit.

jgstroud:
I had 2 crashes after 10 days.  I haven't analyzed the crash dumps yet to see if it is related to this patch, but leading up to that I had a storm of HK log messages asking for the firmware version.  hundreds per second.  so maybe unrelated.

jgstroud:
> Spoke too soon. I found another one... `lastDoorUpdateAt` in web.cpp needs to be declared 64-bit.

I thought I got that one 🤔

jgstroud:
@dkerr64 pretty sure this change has introduced some instabilities.  It needs some rethinking before merging.

dkerr64:
ok, that is concerning.  I would not have expected a change from storing a value as 32bit to storing at 64bit to make any difference.

jgstroud:
There are only a few places where I think the rollover isn't handled properly.  So, I think I'll fix that and keep with the 32bit version.  I think we should only use the 64bit for the uptime.  The ESP32 probably better handles 64bit.  But if you are seeing crashes, I wouldn't rule this out.

jameyers4:
Is there any update to this issue please @jgstroud ?

jgstroud:
No.  I haven't had any time to work on it.

jgstroud:
Yes. Please open a new PR.  This is the better approach.  I knew this was the better approach to begin with but I thought I could quickly hack in a fix which was clearly a bad decision for all the reasons you mentioned.

dkerr64:
Closing this as PR #276 has better fix.

PR #276, opened 2025-06-28
Fix millis() rollover bugs with 32-bit arithmetic

frankea:
## Summary
- Fixes millis() rollover issues after ~49.7 days using rollover-safe comparison patterns
- Replaces direct time comparisons with subtraction-based elapsed time checks
- Avoids 64-bit arithmetic overhead that was causing stability issues in PR #262

## Problem
After ~49.7 days of uptime, millis() rolls over from 4294967295 back to 0. Direct comparisons like `if (millis() > timeout)` fail because the timeout value becomes larger than the rolled-over millis() value.

## Solution
Uses the standard embedded pattern for rollover-safe timing:
```cpp
// Instead of: if (millis() > timeout)
// Use: if (millis() - startTime >= duration)
```

This works because unsigned arithmetic wraps correctly even across rollover boundaries.

## Fixed Issues
- **manual_recovery()**: Removed timeout field from ForceRecover struct, uses local static start_time
- **WiFi timeouts**: Fixed gateway ping, soft AP mode, and connection timeout handling
- **Status timeout**: Changed from absolute time to elapsed time comparison
- **Motion/LED timers**: Added signed cast for proper rollover detection
- **Heap checking**: Converted to elapsed time pattern

## Benefits over 64-bit approach
- No performance impact (ESP8266 lacks hardware 64-bit support)
- No additional memory usage
- Eliminates stability concerns from PR #262
- Surgical fix targeting only the problematic comparisons

### Unit Test (Validates the Math)
A unit test is included that proves the rollover-safe timing logic works correctly:

```bash
gcc -o rollover_test rollover_unit_test_fixed.cpp && ./rollover_test
```

**Result**: All rollover timing tests passed.

### Accelerated Rollover Test (Validates Real Device)
Since waiting 49+ days isn't practical, I've created an accelerated test that simulates rollover in 2 minutes:

#### Setup:
1. Add this header file: `src/test_millis.h` (included in PR)
2. Add `#include "test_millis.h"` to `src/ratgdo.cpp` 
3. Add `-D TEST_ROLLOVER` to build_flags
4. Add `log_rollover_status();` call to main loop

#### Test Instructions:
```bash
# Build test version
pio run -e test_rollover

# Flash and monitor
pio run -e test_rollover -t upload
pio device monitor -e test_rollover
```

#### Expected Output:
```
ROLLOVER TEST: millis()=4294847295, original=120000, offset=4294727295
...
*** ROLLOVER OCCURRED ***
ROLLOVER TEST: millis()=10000, original=240000, offset=4294727295
```

#### Test Cases:
**Minutes 0-2 (Pre-rollover):**
- Press wall button 5 times in 3 seconds → should enter WiFi recovery
- Change WiFi settings → should timeout in 30 seconds
- Test motion detection → should clear after 5 seconds
- Reboot device → HomeKit should start after 2 seconds

**Minutes 2-4 (Post-rollover):**
- Repeat all same tests → should work identically
- Verify web interface continues working
- Check that no features hang or malfunction

#### Success Criteria:
- All timeouts work correctly before rollover
- Rollover occurs without crashes or hangs
- All timeouts work correctly after rollover  
- No infinite waits or premature timeouts
- Device functions normally throughout test

### Alternative: Mathematical Verification
The fix can also be verified by code inspection. The pattern `(current_time - start_time >= duration)` is mathematically proven to work across unsigned integer rollover because:

1. Unsigned arithmetic wraps correctly in C/C++
2. `current_time - start_time` gives correct elapsed time even when `current_time < start_time` due to rollover
3. This is the standard embedded systems approach for handling timer rollover

## Files for Testing
All test files are included in this PR:
- `rollover_unit_test_fixed.cpp` - Unit test 
- `src/test_millis.h` - Rollover simulation header
- `rollover_test_instructions.md` - Detailed test guide

This provides a complete solution to the testing challenge mentioned in #261: "There is no way to artificially test this as far as I can tell" - now there is.

Resolves the issues identified in #261 without the complexity and instability of the 64-bit conversion.

### DIFF
--- a/rollover_test_instructions.md
+++ b/rollover_test_instructions.md
@@ -1,0 +1,92 @@
+# Accelerated Rollover Test Instructions
+
+## Setup
+
+1. **Apply the test files:**
+   ```bash
+   # The test_millis.h file is already created
+   # Apply the patch to ratgdo.cpp:
+   patch -p1 < apply_rollover_test.patch
+   ```
+
+2. **Build with test flags:**
+   ```bash
+   # Using the test environment
+   pio run -e test_rollover
+   
+   # Or manually add to your build_flags in platformio.ini:
+   # -D TEST_ROLLOVER
+   ```
+
+3. **Flash and monitor:**
+   ```bash
+   pio run -e test_rollover -t upload
+   pio device monitor -e test_rollover
+   ```
+
+## What Happens
+
+- Device starts with millis() = ~4294847295 (2 minutes before rollover)
+- After ~2 minutes of runtime, millis() will wrap to 0
+- Serial output shows rollover status every 10 seconds
+- You'll see "*** ROLLOVER OCCURRED! ***" when it happens
+
+## Test Schedule
+
+**Minutes 0-2: Pre-rollover**
+- Test manual recovery (press wall button 5 times quickly)
+- Test motion detection if available
+- Change WiFi settings to test 30-second timeout
+- Verify all features work normally
+
+**Minutes 2-4: Post-rollover**  
+- Repeat all the same tests
+- Verify timing still works correctly
+- Check that no features hang or malfunction
+
+## Expected Serial Output
+
+```
+ROLLOVER TEST: millis()=4294847295, original=120000, offset=4294727295
+ROLLOVER TEST: millis()=4294857295, original=130000, offset=4294727295
+...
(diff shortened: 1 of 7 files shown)


frankea:
Closing this PR as these changes are already included in PR #273 which contains all the comprehensive fixes together.

Issue #354, opened 2026-09-12
ESP8266 v2.1.4: flash sector 0 (eboot) overwritten with zeros after a power outage, device won't boot


smudgeface:
### Summary

My ratgdo v2.5i on homekit-ratgdo v2.1.4 had been running reliably. After a power outage it stopped booting. The first 2,248 bytes of flash sector 0, where the eboot bootloader lives, had been programmed to `0x00` and the sector had never been erased. The rest of flash was intact, including the app image, config and HomeKit pairing. Rewriting just sector 0 from the release binary fixed it with no reconfiguration or re-pairing.

I'm reporting it because the damage pattern looks like a software write to address 0 rather than random corruption. I couldn't find a code path in v2.1.4 that does this, though, so it may be brownout-related. Details are below in case they help anyone else or ring a bell.

### Environment

- Hardware: ratgdo v2.5i, ESP8266EX, 26 MHz crystal, CH340 USB-serial
- Flash: 4MB, JEDEC manufacturer `0x5E`, device `0x4016`
- Firmware: homekit-ratgdo v2.1.4 (`homekit-ratgdo-v2.1.4.bin`, md5 `fc1c3917f1fd35b14e2f1344b2dd2e3d`)
- Opener: Security+ 2.0
- A staged copy of the full v2.1.4 image was still present at `0x13F000`, so v2.1.4 was presumably installed by OTA

### Symptoms

After power returned, the blue LED blinked once at power-up and never again. The device didn't join WiFi and didn't start the soft AP. The ROM boot log at 74880 baud showed:

```
 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

ets_main.c
```

Nothing followed at 115200. Boot mode `(3,6)` is a normal flash boot, so the strapping pins were fine. The ROM didn't find a valid image at `0x0`.

### Flash analysis

I dumped all 4MB with `esptool.py read_flash` before changing anything and compared it with the v2.1.4 release binary:

| Region | Result |
|---|---|
| `0x1000` – end of image | Byte-for-byte identical to the release |
| Staged OTA copy at `0x13F000` | Byte-for-byte identical to the release, including its eboot sector |
| HomeKit/EEPROM sector `0x3FB000`, LittleFS | Intact; config and pairing survived |
| Sector 0, `0x000`–`0x8C7` | **All `0x00`** (2,248 bytes, 8 full pages plus 200 bytes) |
[...]

PR #355, opened 2026-09-15
Skip config writes when nothing changed, and don't delete config before rename (#354)

smudgeface:
Follow-up to #354. This fixes the two config-save problems discussed there, problem 4 (config written on every IP acquisition) and problem 3 (config file deleted before the rename), as two separate commits.

### 1. Only write user config to flash when a setting has changed

On ESP8266, `userSettings::save()` rewrote `user_config` on every call, whether or not anything had changed. `onGotIP()` calls it on every WiFi connection, so every boot and every reconnect caused a flash write.

I looked at why that save is there. Before the config refactor in 304a6e0, `onGotIP()` wrote each IP value to its own file on every connection, and the refactor carried that over as a whole-config write. The saved values (the last DHCP address, shown in the web UI and used when switching to a static IP) are still useful. The write just doesn't need to happen when they haven't changed.

Instead of special-casing `onGotIP()`, this tracks changes in `userSettings`:
- `set()` marks the settings dirty only if the value actually changes. Strings are compared only up to what fits in the buffer, since `strlcpy()` truncates.
- `save()` returns early when nothing is dirty. It still writes if the config file doesn't exist yet, e.g. on first boot or after `erase()`.
- `load()` clears the flag. `save()` clears it only after the rename succeeds, so a failed save is retried next time.

I did it this way because several places call `set()` without saving straight away, and currently rely on a later save to persist the value: the automatic time zone from `get_auto_timezone()`, and the WiFi resets in `set_new_ssid()` when run from the serial console. A fix limited to comparing IP addresses in `onGotIP()` could silently stop persisting those, for example when a new network hands out the same IP. With change tracking, any pending change is still written at the next `save()`, as it is today. The only difference is that saves with nothing to write are skipped, and that now applies to all 18 `ESP8266_SAVE_CONFIG()` call sites, not just `onGotIP()`.

Two side effects to be aware of:
- **New keys.** Previously the got-IP save rewrote the whole file on every boot, which also wrote out any keys a firmware update had added, with their current defaults. Now a key missing from the file stays missing until some setting actually changes, and the device uses the default held in RAM in the meantime. Nothing depends on this today. But if a future release changes a default, a device that has never saved a change will pick up the new default rather than keep the old one.
- **Damaged file.** An existing but empty or corrupt `user_config` is no longer rewritten on the next WiFi connection. The device runs on defaults either way, and the next real change writes the complete file. The second commit also makes this state much less likely.

### 2. Don't delete the config file before renaming the new one

`save()` writes `user_config.tmp`, then removes `user_config` before renaming the temp file over it. A power loss between the remove and the rename leaves no config file at all, and `load()` doesn't look at `.tmp`, so the device comes up with defaults.

The remove isn't needed. The littlefs (v2.5) bundled with the ESP8266 core handles a rename onto an existing file by replacing it in the same metadata commit (`lfs_rawrename()` in `lfs.c`), and `LittleFSImpl::rename()` calls `lfs_rename()` directly. Dropping the `exists()`/`remove()` means `user_config` always contains either the old or the new contents. If the rename itself fails, e.g. on an I/O error or a full filesystem, the existing file is left untouched instead of already deleted.

### ESP32

see my comments on #354 for my analysis of how ESP32 is affected. In short: it isn’t. 

### Testing

- With both commits, flash use grows by 340 bytes and RAM is unchanged.
- `pio test -e native --filter test_core` passes 7/7, before and after.
- The existing native tests don't cover `userSettings`, since it depends on LittleFS and the Arduino core, so I haven't added a unit test.
- Tested on my ratgdo v2.5i, updated OTA from v2.1.4 to a `ratgdo_esp8266_hV25` build of this branch (reports 2.2.5), with the log level at Debug:
  - After boot, `onGotIP()` logs `User configuration unchanged, not writing to file: user_config` instead of rewriting the file.
  - Changing a setting (log level 3 → 4) logs `Writing user configuration to file`. After a reboot the new value is still there, so the rename over the existing file works.
  - Changing it back (4 → 3) and rebooting also persisted.
  - No errors in the boot log. All other settings and the HomeKit pairing were unchanged across the update and the reboots.

Unrelated, but I noticed while running the CI steps locally: `pio test -e native --filter test_packet` and `--filter test_reader` in `test.yml` match no test cases, so those two steps currently test nothing.


dkerr64:
Thank you, it makes sense to confine the changes to config.cpp/config.h.  I see no reason not to merge this.

Thanks

